PR #4, opened 2023-03-08
Feature/tokenomics module

cosmic-vagabond:


proto/elys/tokenomics/time_based_inflation.proto
```diff
@@ -0,0 +1,15 @@
+syntax = "proto3";
+package elysnetwork.elys.tokenomics;
+
+import "elys/tokenomics/inflation_entry.proto"; 
+
+option go_package = "github.com/elys-network/elys/x/tokenomics/types";
+
+message TimeBasedInflation {
+  uint64 startBlockHeight = 1; 
+  uint64 endBlockHeight = 2;
```
AustinoBombino, line 10: @cosmic-vagabond If we're going to be handing inflation through the epochs module won't we need a time duration rather than an endblock?

proto/elys/tokenomics/airdrop.proto
```diff
@@ -0,0 +1,11 @@
+syntax = "proto3";
+package elysnetwork.elys.tokenomics;
+
+option go_package = "github.com/elys-network/elys/x/tokenomics/types";
+
+message Airdrop {
+  string intent = 1; 
+  uint64 amount = 2; 
+  string authority = 3;
```
AustinoBombino, line 9: @cosmic-vagabond It would be useful to have descriptions of what all the new data structure fields are. Either in a spec or the readme or in comments somewhere. Just to help with readability so we don't have to search around to see what does what.

proto/elys/tokenomics/time_based_inflation.proto
cosmic-vagabond: @AustinoBombino for this module I was thinking of not relying on the epoch module but rather just use the startBlockHeight to calculate which Yearly inflation values should be used.

proto/elys/tokenomics/airdrop.proto
cosmic-vagabond: @AustinoBombino makes sense I will add some specs document to the code.

cosmic-vagabond:
Hey @AustinoBombino, just letting you know that I'm merging this pull request so it can be included in the `v0.1.0` release. I'll add the documentation updates in a separate pull request for the next release. Thanks!

PR #10, opened 2023-03-29
Oracle v1

jelysn:
* bootstrap oracle module with ignite cli and update ibc-go version to v6


proto/elys/oracle/params.proto
```diff
@@ -0,0 +1,31 @@
...
+  repeated cosmos.base.v1beta1.Coin fee_limit = 6 [
+    (gogoproto.nullable) = false,
+    (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coins"
+  ];
+  uint64 prepare_gas = 7;
+  uint64 execute_gas = 8;
+  string client_id = 9 [ (gogoproto.customname) = "ClientID" ];
+  string band_epoch = 10;
+  uint64 price_expiry_time = 11;
+  string module_admin = 12;
```
cosmic-vagabond, line 30: @jelysn can you explain the purpose of module_admin

x/oracle/genesis_test.go
```diff
@@ -0,0 +1,58 @@
...
+// 	nullify.Fill(&genesisState)
+// 	nullify.Fill(got)
+
+// 	require.Equal(t, genesisState.PortId, got.PortId)
+
+// 	require.ElementsMatch(t, genesisState.AssetInfos, got.AssetInfos)
+// 	require.ElementsMatch(t, genesisState.Prices, got.Prices)
+// require.ElementsMatch(t, genesisState.PriceFeeders, got.PriceFeeders)
+// this line is used by starport scaffolding # genesis/test/assert
+// }
```
cosmic-vagabond, line 58: @jelysn what do we plan to do with those tests?

cosmic-vagabond (approved):
LGTM

x/oracle/genesis_test.go
jelysn: I was planning to use simapp test for most of the cases, and disabled it - will need to recover this test on next PR

proto/elys/oracle/params.proto
jelysn: module admin is the user who has permission to set data feeder. (Others can't do it) 
Since we discussed to not put admin functions on testnet, it will need to be removed when going for testnet.

PR #37, opened 2023-04-12
Fix: Cancel Vest 

AustinoBombino:
Adds canceled vest tokens back to address's uncommitted tokens.

### DIFF
--- a/x/commitment/keeper/msg_server_cancel_vest.go
+++ b/x/commitment/keeper/msg_server_cancel_vest.go
@@ -51,6 +51,20 @@ func (k msgServer) CancelVest(goCtx context.Context, msg *types.MsgCancelVest) (
 		return nil, sdkerrors.Wrapf(types.ErrInsufficientVestingTokens, "denom: %s, amount: %s", msg.Denom, msg.Amount)
 	}
 
+	// Update the uncommitted tokens amount
+	uncommittedToken, found := commitments.GetUncommittedTokensForDenom(msg.Denom)
+
+	if found {
+		uncommittedToken.Amount = uncommittedToken.Amount.Add(msg.Amount)
+	} else {
+		uncommittedTokens := commitments.GetUncommittedTokens()
+		uncommittedTokens = append(uncommittedTokens, &types.UncommittedTokens{
+			Denom:  msg.Denom,
+			Amount: msg.Amount,
+		})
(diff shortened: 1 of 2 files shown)


x/commitment/keeper/msg_server_cancel_vest.go
```diff
@@ -51,6 +51,20 @@ func (k msgServer) CancelVest(goCtx context.Context, msg *types.MsgCancelVest) (
...
+
+	if found {
+		uncommittedToken.Amount = uncommittedToken.Amount.Add(msg.Amount)
+	} else {
+		uncommittedTokens := commitments.GetUncommittedTokens()
+		uncommittedTokens = append(uncommittedTokens, &types.UncommittedTokens{
+			Denom:  msg.Denom,
+			Amount: msg.Amount,
+		})
+		commitments.UncommittedTokens = uncommittedTokens
```
cosmic-vagabond, line 65: @AustinoBombino I think this line need to apply regardless of the `found` condition correct?

x/commitment/keeper/msg_server_cancel_vest.go
```diff
@@ -51,6 +51,20 @@ func (k msgServer) CancelVest(goCtx context.Context, msg *types.MsgCancelVest) (
 		return nil, sdkerrors.Wrapf(types.ErrInsufficientVestingTokens, "denom: %s, amount: %s", msg.Denom, msg.Amount)
 	}
 
+	// Update the uncommitted tokens amount
```
cosmic-vagabond, line 54: @AustinoBombino just a recommendation, instead of manually incrementing the `i` counter in the `else` branches, one way you can approach that is to have the standard `i++` in the `for` expression and then have a `i--` within the `if vesting.TotalAmount.IsZero() {` branch that would make code visibility easier to understand I believe.

x/commitment/keeper/msg_server_cancel_vest.go
cosmic-vagabond: we basically need a way to set the changes applied at line 58 with `SetCommitments`

cosmic-vagabond: @AustinoBombino I take it back `GetUncommittedTokensForDenom` is returning a pointer so updating it updates the sub item of `commitments`.

cosmic-vagabond (approved):
LGTM

PR #40, opened 2023-04-13
Feat/incentive

kenta-elys:
- Scaffold incentive module
- Declare params of incentive module - LPIncentive and StakeIncentive
- Add epoch hooks in incentive module
- Add staking hooks in incentive module


cosmic-vagabond (approved):
LGTM

PR #46, opened 2023-04-18
feat: 🎸 move from tendermint to cometbft

cosmic-vagabond:


app/setup_handlers.go
```diff
@@ -30,7 +30,7 @@ func loadUpgradeStore(app *ElysApp) {
 
 	if shouldLoadUpgradeStore(app, upgradeInfo) {
 		storeUpgrades := storetypes.StoreUpgrades{
-			// Added: []string{},
+			Added: []string{"incentive"},
```
jelysn, line 33: I am not sure why this upgrade handler stuff ("incentive") puit on tendermint -> cometbft upgrade

config.yml
```diff
@@ -4,11 +4,6 @@ accounts:
     coins:
       - 9000000000000000uelys
       - 100000000ueden
-    mnemonic: olympic slide park figure frost benefit deer reform fly pull price airport submit monitor silk insect uphold convince pupil project ignore roof warfare slight
-  - name: bob
-    coins:
-      - 9000000000000000uelys
-    mnemonic: lucky toe fee soul alley subway rhythm weather danger convince huge rug tennis churn skate return finish service vivid route cabin idea brush wear
```
jelysn, line 11: Bob account was not useful?

app/setup_handlers.go
cosmic-vagabond: it is because it was missing on v0.3.1, when we created the new module incentive we need to create the module kv store the first time, it is part of the migration init process

jelysn: Okay

config.yml
jelysn: Okay

jelysn (approved):
LGTM

PR #51, opened 2023-04-20
Feat/incentive

kenta-elys:
1. Declare Edenboost denom in incentive keys.go.
2. Implement incentive module of uncommitted tokens calculation.
- Update total incentive params ( Total elys staked, Total eden committed, Total eden boost committed)
- Calculate a new uncommitted eden tokens based on its share. Calculate individual delegated amount per commitment owner.
- Calculate a new uncommitted eden boost tokens based on 100% APR of Elys staked and Eden committed.
- Calculate epoch counts for chosen incentive param.

3. Add a Standardstakingtoken function in commitment module to initiate commitment entry for standard staking
4. Declare genesis incentive params in config.yml


x/commitment/keeper/keeper.go
```diff
@@ -6,18 +6,37 @@ import (
...
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 	paramtypes "github.com/cosmos/cosmos-sdk/x/params/types"
 	"github.com/tendermint/tendermint/libs/log"
 
 	"github.com/elys-network/elys/x/commitment/types"
 )
 
+// Interface declearation
+type CommitmentKeeperI interface {
```
jelysn, line 17: Could you put this interface on  types/expected_keepers.go

x/commitment/keeper/msg_server_deposit_tokens.go
```diff
@@ -56,5 +56,18 @@ func (k msgServer) DepositTokens(goCtx context.Context, msg *types.MsgDepositTok
 	// Update the commitments
 	k.SetCommitments(ctx, commitments)
 
+	// Emit Hook commitment changed
+	k.HookCommitmentChanged(ctx, msg.Creator, sdk.NewCoin(msg.Denom, msg.Amount))
```
jelysn, line 60: Can you change the name to `OnCommitmentChange` or `AfterCommitmentChange` - normal hook names look like that.

x/commitment/keeper/keeper.go
kenta-elys: This is interface declaration not reference.

x/commitment/keeper/msg_server_deposit_tokens.go
kenta-elys: Okay noted.

x/incentive/keeper/keeper.go
```diff
@@ -39,10 +45,267 @@ func NewKeeper(
...
+			delegatedAmt.Add(amt)
+			return false
+		},
+	)
+
+	return delegatedAmt
+}
+
+// Find out active incentive params
+func (k Keeper) FindProperIncentiveParm(ctx sdk.Context, epochIdentifier string) (bool, types.IncentiveInfo, types.IncentiveInfo) {
```
jelysn, line 132: Can you update the name to full version? `FindProperIncentiveParam` or `GetProperIncentiveParam`

jelysn: The name dooes not make sense I think.

x/incentive/keeper/keeper.go
```diff
@@ -39,10 +45,267 @@ func NewKeeper(
...
+			creator := commitments.Creator
+
+			// Calculate delegated amount per delegator
+			delegatedAmt := k.CalculateDelegatedAmount(ctx, creator)
+
+			// Calculate new uncommitted Eden tokens for LP, staker, and Eden token holders
+			new_uncommitted_eden_tokens := k.CalculateNewUncommittedEdenTokens(ctx, delegatedAmt, commitments, stakeIncentive, lpIncentive)
+
+			// Calculate new uncommitted Eden-Boost tokens for staker and Eden token holders
+			new_uncommitted_eden_boost_tokens := k.CalculateNewUncommittedEdenBoostTokens(ctx, delegatedAmt, commitments, epochIdentifier, stakeIncentive, lpIncentive)
```
jelysn, line 195: I think golang does not use snake case variable name but will need to use camel case mostly. `newUncommittedEdenBoostTokens`

x/incentive/keeper/keeper.go
kenta-elys: Ok

proto/elys/incentive/params.proto
```diff
@@ -9,6 +9,6 @@ option go_package = "github.com/elys-network/elys/x/incentive/types";
 // Params defines the parameters for the module.
 message Params {
   option (gogoproto.goproto_stringer) = false;
-	repeated IncentiveInfo LPIncentives = 1 [(gogoproto.nullable) = false];
-	repeated IncentiveInfo StakeIncentives = 2 [(gogoproto.nullable) = false];
+	repeated IncentiveInfo lp_incentives = 1 [(gogoproto.nullable) = false];
+	repeated IncentiveInfo stake_incentives = 2 [(gogoproto.nullable) = false];
```
AustinoBombino, line 13: @kenta-elys Did we decide not to use the x/tokenomics InflationEntry type for this?

kenta-elys: I have a thought to fetch only amount data from tokenomic module. I could change to use tokenomic inflationentry no problem. Could you please focus on checking the formula that i have used to calculate the uncommitted token amount. And also the eden and eden boost token handling?

x/incentive/keeper/keeper.go
```diff
@@ -39,10 +45,267 @@ func NewKeeper(
...
+
+			// Eden boost
+			if c.Denom == types.EdenB {
+				k.tci.TotalEdenBoostCommitted = k.tci.TotalEdenCommitted.Add(c.Amount)
+			}
+		}
+
+		return false
+	})
+}
```
AustinoBombino, line 84: @kenta-elys We may be able to improve this by tracking the totalcommitted amounts for each denom in the commitment module and updating them when changes are made. That way we don't have to iterate all commitments every epoch.

x/incentive/keeper/keeper.go
```diff
@@ -39,10 +45,267 @@ func NewKeeper(
...
+	// Recalculate total committed info
+	k.UpdateTotalCommitmentInfo(ctx, epochIdentifier)
+
+	// Iterate all delegations for the specified delegator
+	// Process to increase uncomitted token amount of Eden & Eden boost
+	k.cmk.IterateCommitments(
+		ctx, epochIdentifier,
+		func(commitments ctypes.Commitments) bool {
+			// Find out active incentive params
+			foundIncentive, stakeIncentive, lpIncentive := k.FindProperIncentiveParm(ctx, epochIdentifier)
```
AustinoBombino, line 178: @kenta-elys We can pass these Incentive objects in through the hook so that they don't have to be found twice.

x/incentive/keeper/hooks_epoch.go
```diff
@@ -6,10 +6,22 @@ import (
...
+	// Find out incentive param using epochIdentifier and current block timestamp
+	foundIncentive, _, _ := k.FindProperIncentiveParm(ctx, epochIdentifier)
+
+	// If there is no incentive available with the current epoch and timestamp,
+	if !foundIncentive {
+		return
+	}
+
+	// Update uncommitted token amount
+	k.UpdateUncommittedTokens(ctx, epochIdentifier)
```
AustinoBombino, line 24: Pass the incentives in here so they don't need to be retrieved again

x/incentive/keeper/keeper.go
AustinoBombino, line 132: We could also enforce that the 0 index of the incentive objects are always the current incentive period. When they've ended we can remove them and shift the arrays. Then all we have to do is check that they 0 index is within the time period, and not loop through the arrays.

x/incentive/keeper/keeper.go
```diff
@@ -39,10 +45,267 @@ func NewKeeper(
...
+func (k Keeper) CalculateTotalShareOfStaking(amount sdk.Int) sdk.Dec {
+	// Total statked = Elys staked + Eden Committed + Eden boost Committed
+	totalStaked := k.tci.TotalElysBonded.Add(k.tci.TotalEdenCommitted).Add(k.tci.TotalEdenBoostCommitted)
+
+	// Share = Amount / Total Staked
+	return sdk.NewDecFromInt(amount).QuoInt(totalStaked)
+}
+
+// Calculate the delegated amount
+func (k Keeper) CalculateDelegatedAmount(ctx sdk.Context, delegator string) sdk.Int {
```
AustinoBombino, line 96: @kenta-elys I think we can just use GetDelegatorBonded here:

`// GetDelegatorBonded returs the total amount a delegator has bonded.
func (k Keeper) GetDelegatorBonded(ctx sdk.Context, delegator sdk.AccAddress) math.Int {`

proto/elys/incentive/incentive.proto
```diff
@@ -11,10 +11,10 @@ message IncentiveInfo {
...
         [ (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int", (gogoproto.nullable) = false, (gogoproto.moretags) = "yaml:\"amount\""];
     // epoch identifier
-	string epochIdentifier = 2;
+	string epoch_identifier = 2;
     // start_time of the distribution
     google.protobuf.Timestamp start_time = 3 
         [(gogoproto.stdtime) = true, (gogoproto.nullable) = false, (gogoproto.moretags) = "yaml:\"start_time\""];
     // distribution duration
-	int64 numEpochs = 4;
+	int64 num_epochs = 4;
```
AustinoBombino, line 19: @kenta-elys We should add a current_epoch so we can track where we are in the period easier.

AustinoBombino: You can look at the x/commitment vesting functions to see how it's used there

x/incentive/keeper/keeper.go
```diff
@@ -39,10 +45,267 @@ func NewKeeper(
...
+		// Eden committed
+		if c.Denom == types.Eden {
+			edenCommitted = c.Amount
+		}
+
+		// Eden boost committed
+		if c.Denom == types.EdenB {
+			edenBoostCommitted = c.Amount
+		}
+	}
```
AustinoBombino, line 223: @kenta-elys there are functions for this in the commitment type:

```
func (c *Commitments) GetUncommittedAmountForDenom(denom string) sdk.Int {
	for _, token := range c.UncommittedTokens {
		if token.Denom == denom {
			return token.Amount
		}
	}
	return sdk.NewInt(0)
}

func (c *Commitments) GetCommittedAmountForDenom(denom string) sdk.Int {
	for _, token := range c.CommittedTokens {
		if token.Denom == denom {
			return token.Amount
		}
	}
	return sdk.NewInt(0)
}
```

x/incentive/keeper/keeper.go
```diff
@@ -39,10 +45,267 @@ func NewKeeper(
...
+}
+
+// Calculate total share of staking
+func (k Keeper) CalculateTotalShareOfStaking(amount sdk.Int) sdk.Dec {
+	// Total statked = Elys staked + Eden Committed + Eden boost Committed
+	totalStaked := k.tci.TotalElysBonded.Add(k.tci.TotalEdenCommitted).Add(k.tci.TotalEdenBoostCommitted)
+
+	// Share = Amount / Total Staked
+	return sdk.NewDecFromInt(amount).QuoInt(totalStaked)
+}
```
AustinoBombino, line 93: Looks correct to me, we can double check with dream

x/incentive/keeper/keeper.go
```diff
@@ -39,10 +45,267 @@ func NewKeeper(
...
+			edenBoostCommitted = c.Amount
+		}
+	}
+
+	// compute eden reward based on above and param factors for each
+	totalEdenCommittedByStake := delegatedAmt.Add(edenCommitted).Add(edenBoostCommitted).Add(edenCommittedByLP)
+	stakeShare := k.CalculateTotalShareOfStaking(totalEdenCommittedByStake)
+
+	// Calculate eden amount per epoch
+	edenAmountPerEpoch := stakeIncentive.Amount.Quo(sdk.NewInt(stakeIncentive.NumEpochs))
```
AustinoBombino, line 230: @kenta-elys we can probably compute this earlier in the function chain because it will be the same for every commitment object. we don't need to recompute it for every object.

AustinoBombino: We could probably even just change the value stored in the incentive param object to be the amount per epoch so that we don't need to compute it at all.

x/incentive/keeper/keeper.go
```diff
@@ -39,10 +45,267 @@ func NewKeeper(
...
+// Calculate new Eden-Boost token amounts based on the given conditions and user's current uncommitted token balance
+func (k Keeper) CalculateNewUncommittedEdenBoostTokens(ctx sdk.Context, delegatedAmt sdk.Int, commitments ctypes.Commitments, epochIdentifier string, stakeIncentive types.IncentiveInfo, lpIncentive types.IncentiveInfo) sdk.Int {
+	// Get eden commitments
+	edenCommitted := sdk.ZeroInt()
+	for _, c := range commitments.CommittedTokens {
+		// Eden committed
+		if c.Denom == types.Eden {
+			edenCommitted = c.Amount
+		}
+	}
```
AustinoBombino, line 261: there is a type function for this

x/incentive/keeper/keeper.go
```diff
@@ -39,10 +45,267 @@ func NewKeeper(
...
+
+	// Gompute eden reward based on above and param factors for each
+	totalEden := delegatedAmt.Add(edenCommitted)
+
+	// Calculate 100% APR for eden boost
+	epochNumsPerYear := k.CalculateEpochCountsPerYear(epochIdentifier)
+
+	return totalEden.Quo(sdk.NewInt(epochNumsPerYear))
```
AustinoBombino, line 269: This looks correct based on my understanding of the requirements, but will clarify with the product guys.

x/incentive/keeper/keeper.go
```diff
@@ -39,10 +45,267 @@ func NewKeeper(
...
+			delegatedAmt := k.CalculateDelegatedAmount(ctx, creator)
+
+			// Calculate new uncommitted Eden tokens for LP, staker, and Eden token holders
+			new_uncommitted_eden_tokens := k.CalculateNewUncommittedEdenTokens(ctx, delegatedAmt, commitments, stakeIncentive, lpIncentive)
+
+			// Calculate new uncommitted Eden-Boost tokens for staker and Eden token holders
+			new_uncommitted_eden_boost_tokens := k.CalculateNewUncommittedEdenBoostTokens(ctx, delegatedAmt, commitments, epochIdentifier, stakeIncentive, lpIncentive)
+
+			// Update Commitments with new uncommitted token amounts
+			k.UpdateCommitments(ctx, creator, new_uncommitted_eden_tokens, new_uncommitted_eden_boost_tokens)
```
AustinoBombino, line 198: @kenta-elys We already have the creators commitments so we can just pass them in here

x/incentive/keeper/keeper.go
AustinoBombino: @kenta-elys The calculation here is correct, we just need to add a parameter for the APR percentage and include that in the computation. The case we're talking about here would just have it at 100% but we need to be able to change it in the future.

x/incentive/keeper/keeper.go
kenta-elys: Yes I could see this is inefficient. But If I should have to pass it as parameters, commitment module should have reference to incentive module while incentive module also refers it. And this causes an error in golang.

x/incentive/keeper/hooks_epoch.go
kenta-elys: Yep, I couldn't do it. In order to do this, I should update parameter of IterateCommitment function inside commitment module to include incentive parameters. But we already imported commitment module from incentive module so I couldn't import inventive from commitment module.

x/incentive/keeper/keeper.go
kenta-elys: Gotcha. I would make changes to epoch hook interface inside incentive module.

proto/elys/incentive/incentive.proto
kenta-elys: Yep, got it.

x/incentive/keeper/keeper.go
kenta-elys: Got it.

x/incentive/keeper/keeper.go
kenta-elys: Okay

x/incentive/keeper/keeper.go
kenta-elys: totalEdenCommittedByStake is differs from commitment. It's not the number for entire one, its the number of the committed amount for each commitment.

kenta-elys: > We could probably even just change the value stored in the incentive param object to be the amount per epoch so that we don't need to compute it at all.

Dream told me about non-linear number for the rewards amount. And it should be implemented inside tokenomic module and incentive module should fetch amount from it.

x/incentive/keeper/keeper.go
kenta-elys: Got it.

x/incentive/keeper/keeper.go
kenta-elys: Okay got it.

x/incentive/keeper/keeper.go
kenta-elys: Okay got it.

x/incentive/keeper/keeper.go
kenta-elys: Got it.

x/incentive/keeper/keeper.go
kenta-elys: I can't use this as it isn't included in staking keeper interface.

x/incentive/keeper/hooks_epoch.go
kenta-elys: Fixed

x/incentive/keeper/keeper.go
kenta-elys: Fixed.

x/incentive/keeper/keeper.go
kenta-elys: Fixed

x/incentive/keeper/keeper.go
kenta-elys: Fixed

x/incentive/keeper/keeper.go
AustinoBombino: @kenta-elys Are you sure? I'm seeing it exposed as a keeper function in delegation.go

https://github.com/cosmos/cosmos-sdk/blob/f71df80e93bffbf7ce5fbd519c6154a2ee9f991b/x/staking/keeper/delegation.go#L74

x/incentive/keeper/keeper.go
AustinoBombino: @kenta-elys 

> here the 'totalEdenCommittedByStake' differs from commitment. It's not the number for entire one, its the number of the committed amount for each commitment.
I was referring to the `edenAmountPerEpoch`, not the `totalEdenCommittedByStake`

x/incentive/keeper/keeper.go
kenta-elys: Done, fixed.

proto/elys/incentive/incentive.proto
kenta-elys: Done, fixed.

proto/elys/incentive/params.proto
cosmic-vagabond: this change will be approached in a later ticket cc @kenta-elys @AustinoBombino

cosmic-vagabond (approved):
LGTM, final formula based on tokenomics module will be handled in a new PR

PR #55, opened 2023-04-20
Commitment Module Denom Checks

AustinoBombino:
Add checks for denoms that can be committed and withdrawn from the commitment module. 

Ex: 
- Eden can be committed, Elys cannot
- Eden-Boost can be committed but not withdrawn


x/commitment/keeper/msg_server_commit_tokens.go
```diff
@@ -5,12 +5,22 @@ import (
...
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	aptypes "github.com/elys-network/elys/x/assetprofile/types"
 	"github.com/elys-network/elys/x/commitment/types"
 )
 
 func (k msgServer) CommitTokens(goCtx context.Context, msg *types.MsgCommitTokens) (*types.MsgCommitTokensResponse, error) {
 	ctx := sdk.UnwrapSDKContext(goCtx)
 
+	assetProfile, found := k.apKeeper.GetEntry(ctx, msg.Denom)
+	if !found {
```
kenta-elys, line 16: I couldn't find eden boost in asset profile genesis list. How can we commit it as it isn't listed. It

PR #75, opened 2023-05-11
fix: 🐛 disable incentive module hook causing panic

cosmic-vagabond:
due to non initiated params


kenta-elys (approved):
Apologize! Yes I found yesterday this issue through unit testing!

PR #81, opened 2023-05-17
feat: 🎸 add create pool message

cosmic-vagabond:


proto/elys/amm/tx.proto
```diff
@@ -1,7 +1,25 @@
...
\ No newline at end of file
+service Msg {
+  rpc CreatePool (MsgCreatePool) returns (MsgCreatePoolResponse);
+}
+message MsgCreatePool {
+           string                   creator         = 1;
+  repeated cosmos.base.v1beta1.Coin weights         = 2 [(gogoproto.nullable) = false];
+  repeated cosmos.base.v1beta1.Coin initialDeposit  = 3 [(gogoproto.nullable) = false];
+           uint64                   swapFee         = 4;
+           uint64                   exitFee         = 5;
```
jelysn, line 19: Are we using uint64 value instead of sdk.Dec?

cosmic-vagabond: those are percentage values so 0 to 100

jelysn (approved):

PR #82, opened 2023-05-17
Feat/join pool message

cosmic-vagabond:


proto/elys/amm/tx.proto
```diff
@@ -1,7 +1,35 @@
...
+  repeated uint64                   scallingFactors = 7;
+}
+
+message MsgCreatePoolResponse {}
+
+message MsgJoinPool {
+           string                   creator        = 1;
+           uint64                   poolId         = 2;
+  repeated cosmos.base.v1beta1.Coin maxAmountsIn   = 3 [(gogoproto.nullable) = false];
+           string                   shareAmountOut = 4;
```
jelysn, line 31: I think we would require type conversion to sdk.Int

cosmic-vagabond: Yes that's coming in next PR

jelysn (approved):

PR #83, opened 2023-05-17
Feat/exit pool message

cosmic-vagabond:


proto/elys/amm/tx.proto
```diff
@@ -1,7 +1,45 @@
...
+           string                   shareAmountOut = 4;
+}
+
+message MsgJoinPoolResponse {}
+
+message MsgExitPool {
+           string                   creator       = 1;
+           uint64                   poolId        = 2;
+  repeated cosmos.base.v1beta1.Coin maxAmountsOut = 3 [(gogoproto.nullable) = false];
+           string                   shareAmountIn = 4;
```
jelysn, line 41: It would require conversion to sdk.Int

cosmic-vagabond: same here that will be defined in next PR

jelysn (approved):

PR #87, opened 2023-05-22
fix: 🐛 remove unecessary pool parameter for balancer

cosmic-vagabond:


cosmic-vagabond:
> So we won't need [future-governor] [scalling-factors] and smooth weighting?

that's correct none of those params are needed as we are only implementing the balancer pool model

PR #89, opened 2023-05-22
Feat/min self delegation

kenta-elys:
- Add a criteria to check if the validator's min self delegation amount equals to param set ( 1 elys as default)


cosmic-vagabond:
deprecated by https://github.com/elys-network/elys/pull/92

PR #90, opened 2023-05-22
feat: 🎸 create pool message

cosmic-vagabond:


x/amm/types/constants.go
```diff
@@ -0,0 +1,23 @@
+package types
+
+import (
+	"cosmossdk.io/math"
+)
+
+const (
+	OneShareExponent = 18
```
kenta-elys, line 8: Our token is 18 decimals?

x/amm/keeper/keeper_create_pool.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	// Run the initialization logic.
+	if err := k.InitializePool(ctx, pool, sender); err != nil {
+		return 0, err
+	}
+
+	// // Send initial liquidity to the pool's address.
+	// initialPoolLiquidity := msg.InitialLiquidity()
+	// err = k.bankKeeper.SendCoins(ctx, sender, pool.GetAddress(), initialPoolLiquidity)
```
kenta-elys, line 58: Why this was commented? Won't receive user deposits?

proto/elys/amm/pool.proto
```diff
@@ -5,15 +5,20 @@ option go_package = "github.com/elys-network/elys/x/amm/types";
 import "elys/amm/pool_params.proto"; 
 import "elys/amm/pool_asset.proto"; 
 import "gogoproto/gogo.proto"; 
-import "cosmos/base/v1beta1/coin.proto"; 
+import "cosmos/base/v1beta1/coin.proto";
+import "cosmos_proto/cosmos.proto";
 
 message Pool {
```
kenta-elys, line 11: I thought that we would have multiplier per pool. Could you clarify with Dream about this?

jelysn (approved):

x/amm/keeper/keeper_create_pool.go
```diff
@@ -0,0 +1,65 @@
...
+// an error upon failure. The pool creation fee is used to fund the community
+// pool. It will create a dedicated module account for the pool and sends the
+// initial liquidity to the created module account.
+//
+// After the initial liquidity is sent to the pool's account, this function calls an
+// InitializePool function from the source module. That module is responsible for:
+// - saving the pool into its own state
+// - Minting LP shares to pool creator
+// - Setting metadata for the shares
+func (k Keeper) CreatePool(ctx sdk.Context, msg *types.MsgCreatePool) (uint64, error) {
```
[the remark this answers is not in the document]
kenta-elys: Don't we need to check if we already have the same pool - msgCreatePool param? Each time we should create the new one?

x/amm/types/set_initial_pool_assets.go
```diff
@@ -0,0 +1,57 @@
...
+	for _, asset := range p.PoolAssets {
+		exists[asset.Token.Denom] = true
+	}
+
+	newTotalWeight := p.TotalWeight
+	scaledPoolAssets := make([]*PoolAsset, 0, len(PoolAssets))
+
+	// TODO: Refactor this into PoolAsset.validate()
+	for _, asset := range PoolAssets {
+		if asset.Token.Amount.LTE(sdk.ZeroInt()) {
```
kenta-elys, line 28: asset.Token won't be available as it were not set before.

x/amm/types/create_pool.go
```diff
@@ -0,0 +1,15 @@
...
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+func (msg *MsgCreatePool) CreatePool(ctx sdk.Context, poolID uint64) (*Pool, error) {
+	poolParams := PoolParams{
+		SwapFee: msg.SwapFee,
+		ExitFee: msg.ExitFee,
+	}
+	poolAssets := make([]PoolAsset, len(msg.InitialDeposit))
```
kenta-elys, line 12: I think we should initialize the value of poolAssets. Its simply create n records of empty poolAssets. But later on, in setInitialAssets function, asset.Token is being referenced.

x/amm/keeper/initialize_pool.go
```diff
@@ -0,0 +1,54 @@
...
+// - Calls the AfterPoolCreated hook
+func (k Keeper) InitializePool(ctx sdk.Context, pool *types.Pool, sender sdk.AccAddress) (err error) {
+	// Mint the initial pool shares share token to the sender
+	err = k.MintPoolShareToAccount(ctx, pool, sender, pool.GetTotalShares().Amount)
+	if err != nil {
+		return err
+	}
+
+	// Finally, add the share token's meta data to the bank keeper.
+	poolShareBaseDenom := types.GetPoolShareDenom(pool.GetPoolId())
```
kenta-elys, line 25: What is the different between poolsharebase denom and pool display denom. pool share = lp token denom?

x/amm/keeper/keeper_create_pool.go
cosmic-vagabond: the pool liquidity field is only available in the stable swap pool model in osmo, I have to look at why it is not available for the balancer model to have that functionality available, it will be enable in the next PR

PR #95, opened 2023-05-26
Feat/swap logic

cosmic-vagabond:


docs/static/openapi.yml
```diff
@@ -69455,6 +69633,44 @@ definitions:
         type: string
       exitFee:
         type: string
+  elysnetwork.elys.amm.QueryAllDenomLiquidityResponse:
```
jelysn, line 69636: Same here for DenomLiquidity -> Liquidity

proto/elys/amm/denom_liquidity.proto
```diff
@@ -0,0 +1,14 @@
...
+package elysnetwork.elys.amm;
+
+import "gogoproto/gogo.proto";
+import "cosmos/base/v1beta1/coin.proto";
+
+option go_package = "github.com/elys-network/elys/x/amm/types";
+
+message DenomLiquidity {
+  string denom = 1; 
+  string liquidity = 2 [(gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int", (gogoproto.moretags) = "yaml:\"taken_out_min_amount\"", (gogoproto.nullable) = false];
```
jelysn, line 11: more tag name `taken_out_min_amount` looks invalid.
Could we just use sdk.Coin object instead of DenomLiquidity struct? I think it would be better to reduce number of newly introducing structs

proto/elys/amm/swap_route.proto
```diff
@@ -0,0 +1,20 @@
...
+message SwapAmountInRoute {
+  uint64 pool_id = 1 [ (gogoproto.moretags) = "yaml:\"pool_id\"" ];
+  string token_out_denom = 2
+      [ (gogoproto.moretags) = "yaml:\"token_out_denom\"" ];
+}
+
+message SwapAmountOutRoute {
+  uint64 pool_id = 1 [ (gogoproto.moretags) = "yaml:\"pool_id\"" ];
+  string token_in_denom = 2
+      [ (gogoproto.moretags) = "yaml:\"token_out_denom\"" ];
```
jelysn, line 19: Can you remove more tag or name it to be same as the field name? (token_in_denom vs token_out_denom)

proto/elys/amm/tx.proto
```diff
@@ -46,22 +47,24 @@ message MsgExitPool {
...
-  repeated string                   swapRouteDenoms   = 5;
+           string                   sender           = 1;
+           repeated SwapAmountInRoute routes = 2 [ (gogoproto.nullable) = false ];
+           cosmos.base.v1beta1.Coin tokenIn           = 3 [(gogoproto.nullable)   = false                                    ] ;
+           string                   tokenOutMinAmount = 4 [(gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int", (gogoproto.moretags) = "yaml:\"taken_out_min_amount\"", (gogoproto.nullable) = false];
 }
 
-message MsgSwapExactAmountInResponse {}
+message MsgSwapExactAmountInResponse {
+           string                   tokenOutAmount = 1 [(gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int", (gogoproto.moretags) = "yaml:\"taken_out_max_amount\"", (gogoproto.nullable) = false];
```
jelysn, line 57: Inconsistent more tag with field name - tokenOutAmount vs taken_out_max_amount

proto/elys/amm/tx.proto
```diff
@@ -46,22 +47,24 @@ message MsgExitPool {
...
 message MsgSwapExactAmountOut {
-           string                   creator           = 1;
-           cosmos.base.v1beta1.Coin tokenOut          = 2 [(gogoproto.nullable) = false];
-           string                   tokenOutMaxAmount = 3 [(gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Uint", (gogoproto.moretags) = "yaml:\"taken_out_max_amount\"", (gogoproto.nullable) = false];
-  repeated uint64                   swapRoutePoolIds  = 4;
-  repeated string                   swapRouteDenoms   = 5;
+           string                   sender           = 1;
+           repeated SwapAmountOutRoute routes = 2 [ (gogoproto.nullable) = false ];
+           cosmos.base.v1beta1.Coin tokenOut          = 3 [(gogoproto.nullable) = false];
+           string                   tokenInMaxAmount = 4 [(gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int", (gogoproto.moretags) = "yaml:\"taken_out_max_amount\"", (gogoproto.nullable) = false];
```
jelysn, line 64: tokenInMaxAmount vs taken_out_max_amount - I recommend to remove all moretags

proto/elys/amm/tx.proto
```diff
@@ -46,22 +47,24 @@ message MsgExitPool {
...
 }
 
-message MsgSwapExactAmountOutResponse {}
+message MsgSwapExactAmountOutResponse {
+           string                   tokenInAmount = 1 [(gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int", (gogoproto.moretags) = "yaml:\"taken_out_max_amount\"", (gogoproto.nullable) = false];
```
jelysn, line 68: Same for different name in more tag and field name

x/amm/keeper/denom_liquidity_test.go
```diff
@@ -0,0 +1,64 @@
...
+			nullify.Fill(&rst),
+		)
+	}
+}
+func TestDenomLiquidityRemove(t *testing.T) {
+	keeper, ctx := keepertest.AmmKeeper(t)
+	items := createNDenomLiquidity(keeper, ctx, 10)
+	for _, item := range items {
+		keeper.RemoveDenomLiquidity(ctx,
+			item.Denom,
```
jelysn, line 48: Could you update all of similar calls with less number of parameters function call to be inline? (I think it does have lots of not required multiline function calls

x/amm/keeper/is-elys-routed-multihop.go
```diff
@@ -0,0 +1,31 @@
...
+	if inDenom == outDenom {
+		return false
+	}
+	poolIds := route.PoolIds()
+	if poolIds[0] == poolIds[1] {
+		return false
+	}
+
+	// route0Incentivized := k.poolIncentivesKeeper.IsPoolIncentivized(ctx, poolIds[0])
+	// route1Incentivized := k.poolIncentivesKeeper.IsPoolIncentivized(ctx, poolIds[1])
```
jelysn, line 26: What's this commented lines for

x/amm/types/constants.go
```diff
@@ -6,6 +6,8 @@ import (
 
 const (
 	OneShareExponent = 18
+
+	BalancerGasFeeForSwap = 10_000
```
jelysn, line 10: Can you share what this does?

proto/elys/amm/denom_liquidity.proto
cosmic-vagabond: good catch will fix that thanks

x/amm/types/pow.go
```diff
@@ -0,0 +1,56 @@
...
+
+import (
+	"fmt"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// Don't EVER change after initializing
+// TODO: Analyze choice here.
+var powPrecision, _ = sdk.NewDecFromStr("0.00000001")
```
jelysn, line 11: Do we need to have custom pow function? (osmomath that we used to have as separate package does not solve the problem?)

x/parameter/types/proposal.go
```diff
@@ -1,117 +1,117 @@
...
+
+// Implements Proposal Interface
+var _ gov.Content = &ProposalUpdateMinSelfDelegation{}
+
+// ProposalRoute gets the proposal's router key
+func (sup *ProposalUpdateMinSelfDelegation) ProposalRoute() string { return RouterKey }
+
+// ProposalType is "SoftwareUpgrade"
+func (sup *ProposalUpdateMinSelfDelegation) ProposalType() string {
+	return ProposalTypeUpdateMinSelfDelegation
```
jelysn, line 106: Not sure how parameter module is in this PR.

jelysn:
Overall looks good to me. Can you add example CLI commands for swap operation on scripts directory?

x/amm/types/constants.go
cosmic-vagabond: here: https://github.com/elys-network/elys/blob/feat/swap-logic/x/amm/types/apply_swap.go#L7-L8

x/amm/keeper/is-elys-routed-multihop.go
cosmic-vagabond: this is a functionality from the original code that we don't need so I can actually remove that from the code

x/amm/types/pow.go
cosmic-vagabond: We are not utilizing osmomath or osmoutils since they rely on an outdated version of the cosmos SDK, which includes significant changes that would require a substantial amount of time to resolve. Therefore, instead of fixing these two libraries to be compatible with the latest SDK version we use, I opted to selectively extract and incorporate only the necessary functions.

cosmic-vagabond:
> Overall looks good to me. Can you add example CLI commands for swap operation on scripts directory?

Thanks will add documentations as part of next PR

proto/elys/amm/tx.proto
cosmic-vagabond: all the moretags annotations were removed

x/parameter/types/proposal.go
cosmic-vagabond: me neither

x/amm/keeper/is-elys-routed-multihop.go
jelysn: Okay

x/amm/types/constants.go
jelysn: I see - got it.

x/amm/types/pow.go
jelysn: Ok

jelysn (approved):
LGTM

PR #115, opened 2023-06-13
fix: set default param of incentive module in upgrade handler

kenta-elys:
- Set default parameter of incentive module in upgrade handler with the new paramset pair added.

### DIFF
--- a/app/setup_handlers.go
+++ b/app/setup_handlers.go
@@ -6,6 +6,7 @@ import (
 	m "github.com/cosmos/cosmos-sdk/types/module"
 	"github.com/cosmos/cosmos-sdk/version"
 	"github.com/cosmos/cosmos-sdk/x/upgrade/types"
+	incentivemoduletypes "github.com/elys-network/elys/x/incentive/types"
 )
 
 func SetupHandlers(app *ElysApp) {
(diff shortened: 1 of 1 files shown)


app/setup_handlers.go
```diff
@@ -18,7 +19,14 @@ func setUpgradeHandler(app *ElysApp) {
 	app.UpgradeKeeper.SetUpgradeHandler(version.Version, func(ctx sdk.Context, plan types.Plan, vm m.VersionMap) (m.VersionMap, error) {
 		app.Logger().Info("Running upgrade handler for " + version.Version)
 
-		return app.mm.RunMigrations(ctx, app.configurator, vm)
+		// Run migrations
+		verisionMap, err := app.mm.RunMigrations(ctx, app.configurator, vm)
```
jelysn, line 23: Can you early return if err is not nil?

kenta-elys: Got it.

app/setup_handlers.go
```diff
@@ -18,7 +19,14 @@ func setUpgradeHandler(app *ElysApp) {
...
+
+		// Set Params of incentive module with the new paramset added
+		params := incentivemoduletypes.DefaultParams()
+		app.IncentiveKeeper.SetParams(ctx, params)
```
jelysn, line 27: If previous upgrade handlers are done successfully, this process won't be required since RunMigrations is automatically running InitGenesis of not initialized modules and initGenesis would be setting the default params.
https://github.com/elys-network/elys/blob/feat/upgrade-incentive-params/x/incentive/keeper/genesis.go#L12

Probably, set params before RunMigrations?

kenta-elys: Yeah, I am trying to do chain upgrade in my local. Am not sure with that exactly, anyhow reproducing the error at the moment.

cosmic-vagabond: @kenta-elys we should increament the incentive module version and add a migrator function to the module instead

cosmic-vagabond (changes requested):

app/setup_handlers.go
kenta-elys: @cosmic-vagabond 
Could you commit this one? I am not sure how to do it.

PR #124, opened 2023-06-16
feat: commit lp token to commitment module when join, withdraw it when exit

kenta-elys:
- Commit LP token minted to commitment module when create-pool & join-pool.
- Withdraw LP token committed to exiter's wallet when exit-pool.
- [TODO]
  Unit tests on both cases.


cosmic-vagabond (approved):
LGTM

PR #128, opened 2023-06-22
Swap edge case tests & fix panics on swap

jelysn:
- Handle panic case on RouteExactAmountIn && RouteExactAmountOut on multihop routed for zero division
- Fix zero division case on createElysMultihopExpectedSwapOuts
- Update MsgServerSwapExactAmountIn test to handle multihop case and invalid route case
- Update MsgServerSwapExactAmountOut test to handle multihop case and invalid route case


cosmic-vagabond (approved):
LGTM

PR #148, opened 2023-07-12
feat: complete a bash script to submit update pool info gov proposal

kenta-elys:
- Complete bash script of submitting gov proposal for updating pool multiplier information
- Fix minor issues with csv export - added poolId, write empty string for invalid data, remove quotation, fix type conversion issue with typescript and google spread sheet.

### DIFF
--- a/scripts/csv/data.csv
+++ b/scripts/csv/data.csv
@@ -1,0 +1,9 @@
+﻿Non-linear eden allocation
+
+poolId,poolName,multiplier,eden_next_allocation
+0,juno,0.750,
+1,evmos,0.750,
+2,akt,0.750,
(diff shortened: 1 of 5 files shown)


scripts/csv/data.csv
```diff
@@ -0,0 +1,9 @@
+﻿Non-linear eden allocation
```
jelysn, line 1: Is it fine to include such data on csv file?

kenta-elys: Yep, it is just a title. It is being taken into account and propose bash script.

cosmic-vagabond (approved):
LGTM

PR #165, opened 2023-08-07
Add unit test for splitting swap to 100 times to check out amount difference

jelysn:


kenta-elys (commented):
@jelysn 
I could see 100x swap gave 100 less output than 1 time swap. Have we applied any extra penalty in oracle swap logic than slippage and swap fee?
And is it always 100 less output in such testing scenario?
You've set 0 swap fee. Ideally, 0 swap fee will result the same output for the same input amount even if it is 100x swap or 1x swap.

cosmic-vagabond:
adding here @jelysn comment from slack to answer @kenta-elys ’s concern:

> It’s not extra penalty, it’s decimal cut result.

> We will need to refactor dynamic weighted pool FYI based on discussion with Dream, and I think it would be good to merge this unit test PR for now, since this test case is required, the test result will be changed after refactoring.

cosmic-vagabond (approved):

PR #195, opened 2023-09-19
IBC Transfer middleware

jelysn:


mars.yml
```diff
@@ -0,0 +1,403 @@
+version: 1
```
cosmic-vagabond, line 1: @jelysn can we move this fine within the script folder where the actual script using this file lives?

app/app.go
```diff
@@ -113,6 +113,7 @@ import (
 	ibcclientclient "github.com/cosmos/ibc-go/v7/modules/core/02-client/client"
```
cosmic-vagabond, line 113: @jelysn this PR is missing the setup handle logic to register the store key of the new module as well as the ParamKeyTable

cosmic-vagabond (changes requested):

mars.yml
jelysn: I think this might be useful on other features which require multiple networks setup.

app/app.go
jelysn: This module does not require any storage, let me check ParamKeyTable

mars.yml
cosmic-vagabond: @jelysn what I meant is to keep the root of the repo clean, it would be to move this config file within a separate folder as it will used for now only for this purpose

app/app.go
cosmic-vagabond: @jelysn any updates on that?

mars.yml
jelysn: Ok

jelysn: Moved config.yml to transferhook scripts directory

app/app.go
jelysn: Registered ParamKeyTable on upgrade handler - https://github.com/elys-network/elys/pull/195/files#diff-69701b9ad62c06683b76c4fdb5063d2005ef939af85e8a7191e872aad538ba06R87
Btw, it has logic to create store key and pass it even though it's not used - you mean to build upgrade handler on this PR?

jelysn: https://github.com/elys-network/elys/pull/195/files#diff-0f1d2976054440336a576d47a44a37b80cdf6701dd9113012bce0e3c425819b7R452
https://github.com/elys-network/elys/pull/195/files#diff-0f1d2976054440336a576d47a44a37b80cdf6701dd9113012bce0e3c425819b7R826 (Store Key)

cosmic-vagabond: Yes the param key table needs to be set since 0.47 as long as you are using params in your module.

and corret the new module `transferhook` needs to be added to:

```
		storeUpgrades := storetypes.StoreUpgrades{
			Added: []string{
				clocktypes.ModuleName,
			},
		}
```

jelysn: It's added

cosmic-vagabond (approved):

PR #278, opened 2023-11-27
Add custom recipient field for amm swap functions

jelysn:


x/amm/keeper/msg_server_swap_exact_amount_in.go
```diff
@@ -17,9 +17,13 @@ func (k msgServer) SwapExactAmountIn(goCtx context.Context, msg *types.MsgSwapEx
 		return nil, err
 	}
 
+	recipient, err := sdk.AccAddressFromBech32(msg.Recipient)
+	if err != nil {
+		recipient = sender
```
cosmic-vagabond, line 22: @jelysn I would suggest to throw an error that says invalid recipient address and stop the flow if that happens so the user expecting to receive the fund in the receipt address doesn't get confused as it gets send to the original sender address what do you think?

jelysn: This is the case recipient is empty - when recipient is empty it's returning to sender address, here msg.Recipient is an optional field - for the users using amm directly won't configure recipient but only the trade shield contract will use custom recipient.

proto/elys/amm/tx.proto
```diff
@@ -77,11 +77,12 @@ message MsgSwapExactAmountOut {
            cosmos.base.v1beta1.Coin token_out           = 3 [(gogoproto.nullable)   = false                                   ] ;
            string                   token_in_max_amount = 4 [(gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int", (gogoproto.nullable) = false];
            string                   discount   = 5 [(gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec", (gogoproto.nullable) = false];
+           string                   recipient  = 6;
```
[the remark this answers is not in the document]
jelysn: Let me add the changes on swap by denom as well

x/amm/client/cli/tx_swap_by_denom.go
```diff
@@ -20,17 +20,18 @@ const (
...
-		Args:    cobra.ExactArgs(3),
+		Args:    cobra.ExactArgs(4),
 		RunE: func(cmd *cobra.Command, args []string) (err error) {
 			argAmount, err := sdk.ParseCoinNormalized(args[0])
 			if err != nil {
 				return err
 			}
 			argDenomIn := args[1]
 			argDenomOut := args[2]
+			argRecipient := args[3]
```
cosmic-vagabond, line 34: @jelysn can you make the recipient a flag instead of a fixed command argument so that it is optional, thanks

cosmic-vagabond (approved):

PR #304, opened 2023-12-14
Check parameter initialization and fix for wasmbindings

jelysn:


cosmic-vagabond:
already part of another PR

PR #357, opened 2024-02-01
Grammar correction

jasonsopko:
"Lesser than" is incorrect since "lesser" and "than" both imply a comparison, which makes them redundant when used together. "less than" is proper.


cosmic-vagabond (approved):
@jasonsopko good catch thanks

PR #476, opened 2024-04-30
AddCommittedToken was edited to avoid amount duplicate on Lockup

fenriz07:
## Description

We have a problem when the user joins a pool for a second time or higher because the lockups amount is doubled. This causes a desynchronization between the amount and the lockups that allow an unavailable amount to be withdrawn

### What has Changed?

Function _AddCommittedTokens_ was edited to resolve the problem, in line 54 there was a problem because each time that adds a new committed the amount was duplicated.


x/commitment/types/commitments.go
```diff
@@ -51,7 +51,6 @@ func (c *Commitments) AddCommittedTokens(denom string, amount math.Int, unlockTi
 					li++
 					continue
 				} else if lockup.UnlockTimestamp == unlockTime {
-					c.CommittedTokens[i].Lockups[li].Amount = lockup.Amount.Add(amount)
```
Wgil, line 54: This loop is not doing anything without this line. Is it? You could remove the whole for loop if you think it's wrong, right?

Wgil (changes requested):

x/commitment/types/commitments.go
fenriz07: You're right because this condition `lockup.UnlockTimestamp == unlockTime` always will happen because the lockup was added previously.

Wgil (approved):

cosmic-vagabond (approved):

PR #477, opened 2024-05-01
Update upgrade assure docs

Wgil:


scripts/upgrade-assure/README.md
```diff
@@ -1,6 +1,6 @@
 # Localnet Upgrade and Data Migration Guide for ELYS Network
 
-This document offers a detailed framework for upgrading the localnet used in the ELYS network, encompassing version changes, data migrations, and utilizing testnet snapshots.
+This document offers a detailed framework for upgrading the localnet used in the ELYS network, encompassing version changes, data migrations, and utilizing old testnet snapshots in newwer versions for local development purposes.
```
cosmic-vagabond, line 3: @Wgil small typo with `newwer` to `newer`

cosmic-vagabond: @Wgil `old` here might be misleading, instead we are using latest snapshot available for testnet which happens to be using the previous version since it is the last one available in testnet

Wgil: Thanks

cosmic-vagabond (approved):

PR #574, opened 2024-06-18
Fix APR calculation

jelysn:


x/leveragelp/client/cli/query_rewards.go
```diff
@@ -13,7 +13,7 @@ import (
 
 func CmdRewards() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "get-position [address] [position-ids]",
+		Use:   "rewards [address] [position-ids]",
```
Wgil, line 16: Please update below line too:
`Short: "Query position"`

fenriz07 (approved):

PR #624, opened 2024-06-26
amm hook fix when executed through leveragelp

jelysn:


x/masterchef/keeper/hooks_masterchef.go
```diff
@@ -44,8 +44,7 @@ func (k Keeper) UpdateAccPerShare(ctx sdk.Context, poolId uint64, rewardDenom st
 		return
 	}
 	poolRewardInfo.PoolAccRewardPerShare = poolRewardInfo.PoolAccRewardPerShare.Add(
-		math.LegacyNewDecFromInt(amount.Mul(ammtypes.OneShare)).
-			Quo(math.LegacyNewDecFromInt(totalCommit)),
+		math.LegacyNewDecFromInt(amount.Mul(ammtypes.OneShare)).QuoInt(totalCommit),
```
cosmic-vagabond, line 47: @jelysn can you check that `totalCommit` will never be zero to avoid a division by zero panic

cosmic-vagabond (approved):

PR #671, opened 2024-07-24
Adding PowerApproximation function

avkr003:
## Description
It adds a new function `PowerApproximation` which can calculate power on numbers greater than 2 as well.
The previous function `PowApprox` had that limitation.

The previous function didn't work because `term.MulMut(c).MulMut(x).QuoMut(bigK)` becomes too large for sdk.Dec to store as max allowed number of bits are 315 and max value should be  < 2^256

Closes:
* https://github.com/elys-network/issues/issues/669

### DIFF
--- a/x/amm/types/pow_approx.go
+++ b/x/amm/types/pow_approx.go
@@ -65,7 +65,6 @@ func computeLn(x sdk.Dec) sdk.Dec {
 }
 
 // PowerApproximation This uses formula z = exp(b * ComputeLn(a)) for a^b.
-// `terms` increases the accuracy of ComputeLn(a)
 func PowerApproximation(base sdk.Dec, exp sdk.Dec) sdk.Dec {
 	if !base.IsPositive() {
 		panic(fmt.Errorf("base must be greater than 0"))
--- a/x/amm/types/pow_approx.go
+++ b/x/amm/types/pow_approx.go
@@ -1,9 +1,10 @@
 package types
 
 import (
-	"errors"
 	"fmt"
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	"math"
+	"strconv"
 )
 
 func computeExp(x sdk.Dec) (sdk.Dec, error) {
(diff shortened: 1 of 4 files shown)


x/amm/types/pow_approx.go
```diff
@@ -6,6 +6,84 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
+var ln2 = sdk.MustNewDecFromStr("0.693147180559945309")
```
amityadav0, line 9: is it a constant value ?

amityadav0 (commented):
looks good, thanks!
could you please add tests ?
also a note/comment on why previous function should not be used for values greater than 2 ?

amityadav0:
can we remove the old function ?

x/amm/types/pow_approx.go
avkr003: Yes it's Log(2) base e

amityadav0: can we move it to constants.go ?

x/amm/types/pow_approx.go
```diff
@@ -1,98 +1,138 @@
...
-// 0 <= exp < 1.
-func PowApprox(base sdk.Dec, exp sdk.Dec, precision sdk.Dec) sdk.Dec {
-	if !base.IsPositive() {
-		panic(fmt.Errorf("base must be greater than 0"))
-	}
+var (
+	ln2    = sdk.MustNewDecFromStr("0.693147180559945309")
+	invLn2 = sdk.MustNewDecFromStr("1.442695040888963407")
+	exp    = sdk.MustNewDecFromStr("2.718281828459045235")
+)
```
amityadav0, line 13: can we these constants to constants.go too ?

x/amm/types/pow_approx_test.go
```diff
@@ -0,0 +1,34 @@
...
+		libraryValue := sdk.MustNewDecFromStr(fmt.Sprintf("%f", math.Pow(numberF, expF)))
+		err := libraryValue.Sub(value).Abs()
+		assert := err.LTE(precision)
+		if !assert {
+			count++
+		}
+	}
+	lowInaccuracy := count < ((iterations * 2) / 100) // 2% inaccuracy
+	require.True(t, lowInaccuracy)
+
```
amityadav0, line 33: is it possible to add few edge cases too ? thanks

amityadav0 (approved):
LGTM

cosmic-vagabond (approved):

PR #674, opened 2024-07-25
[Leveragelp] Locked balance improvements

fenriz07:
A new query was created to know what tokens are locked for 1 hour in the leverage lp through an address wallet.


x/leveragelp/client/cli/query_committed_tokens_locked.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/elys-network/elys/x/leveragelp/types"
+	"github.com/spf13/cobra"
+)
+
+var _ = strconv.Itoa(0)
+
+func CmdCommittedTokensLocked() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "committed-tokens-locked [address]",
+		Short: "Query committed-tokens-locked",
```
amityadav0, line 17: - Please add something descriptive maybe `This query shows committed tokens for all the position for a address`
- I think we have similar query in commitment module but that one displays for one address only.

fenriz07: Thank you. It was changed.

cosmic-vagabond (approved):

fenriz07:
Result of the query:
image

PR #675, opened 2024-07-26
Update estaking codec

Wgil:
This is just a PR I intend to test on devnet to see if Amino codecs works. I'll revert after


Wgil:
Closing this already.

PR #683, opened 2024-07-31
avkr003/leverageLPFixes

avkr003:
## Description
This fixes few bugs found in Leverage LP and improves checks as well.


x/leveragelp/keeper/position.go
```diff
@@ -97,9 +92,15 @@ func (k Keeper) DestroyPosition(ctx sdk.Context, positionAddress string, id uint
 		if len(stopLossKey) > 0 {
 			store.Delete(stopLossKey)
 		}
-		store.Delete([]byte(old.GetPositionAddress()))
+		store.Delete(old.GetPositionAddress())
```
amityadav0, line 95:
```suggestion
		store.Delete([]byte(old.GetPositionAddress()))
```

amityadav0: I think this should be bytes, as we store key in bytes

amityadav0 (approved):
looks good

amityadav0:
Please fix tests

x/leveragelp/keeper/position.go
avkr003: AccAddress is always in Byte format

avkr003: Though now I am unsure what it's deleting? No prefix key was used. Ideally it should have types.GetPositionKey to delete, but now that is being called in next line

amityadav0: makes sense.
This will be removed completely as we won't be using sorted maps.

avkr003: Shall I remove this code then?

amityadav0: not yet, it will be part of `new-leverage-flow` branch and we will merge to `main` after everything looks good with new flow

cosmic-vagabond (approved):

PR #685, opened 2024-08-01
Fixing close position for non-zero exit fee pool

avkr003:
This PR fixes the closing position for a non-zero exit fee pool. The issue happens because exitFee is not considered when calculating userAmount.

### DIFF
--- a/x/leveragelp/keeper/position.go
+++ b/x/leveragelp/keeper/position.go
@@ -399,7 +399,7 @@ func (k Keeper) GetPositionHealth(ctx sdk.Context, position types.Position) (sdk
 	for _, commitment := range commitments.CommittedTokens {
 		cacheCtx, _ := ctx.CacheContext()
 		cacheCtx = cacheCtx.WithBlockTime(cacheCtx.BlockTime().Add(time.Hour))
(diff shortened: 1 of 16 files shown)


x/leveragelp/keeper/position.go
```diff
@@ -399,7 +399,7 @@ func (k Keeper) GetPositionHealth(ctx sdk.Context, position types.Position) (sdk
 	for _, commitment := range commitments.CommittedTokens {
 		cacheCtx, _ := ctx.CacheContext()
 		cacheCtx = cacheCtx.WithBlockTime(cacheCtx.BlockTime().Add(time.Hour))
-		exitCoins, err := k.amm.ExitPool(cacheCtx, position.GetPositionAddress(), position.AmmPoolId, commitment.Amount, sdk.Coins{}, depositDenom)
+		exitCoins, _, err := k.amm.ExitPool(cacheCtx, position.GetPositionAddress(), position.AmmPoolId, commitment.Amount, sdk.Coins{}, depositDenom)
```
amityadav0, line 402: This one should also consider `exitCoinsAfterExitFee` because net assets are after deducting fees

amityadav0:
have you tested it with latest testnet snapshot ?  is `insufficient funds` resolved ?

cosmic-vagabond (commented):
Lgtm

PR #689, opened 2024-08-02
Avkr003/leverage lp testing

avkr003:
test cases for opening positions and changing method how position health is calculated


x/leveragelp/keeper/position.go
```diff
@@ -387,27 +388,42 @@ func (k Keeper) GetPositionsForAddress(ctx sdk.Context, positionAddress sdk.Addr
...
-			return sdk.ZeroDec(), err
-		}
-		positionVal = positionVal.Add(sdk.NewDecFromInt(exitCoinsAfterExitFee.AmountOf(depositDenom)))
+		leveragedLpAmount = leveragedLpAmount.Add(commitment.Amount.ToLegacyDec())
 	}
 
-	lr := positionVal.Quo(sdk.NewDecFromBigInt(xl.BigInt()))
-	return lr, nil
+	exitValue := leveragedLpAmount.Mul(tvl).Quo(ammPool.TotalShares.Amount.ToLegacyDec())
+	exitValueAfterFee := exitValue.Mul(sdk.OneDec().Sub(ammPool.PoolParams.ExitFee))
```
amityadav0, line 422: I think this is wrong. when we use ExitPool, asset weights and ratios are also considered which is important when exiting the pool with getting one token(USDC in this case).
for example: If pool has 90ELYS and 10USDC and other case 50ELYS 50USDC
The result will be different in both cases.
But if we use updated formula, that will yield same result

amityadav0: Please correct me if I am wrong

x/leveragelp/types/expected_keepers.go
```diff
@@ -79,3 +79,7 @@ type MasterchefKeeper interface {
 	ClaimRewards(ctx sdk.Context, sender sdk.AccAddress, poolIds []uint64, recipient sdk.AccAddress) error
 	UserPoolPendingReward(ctx sdk.Context, user sdk.AccAddress, poolId uint64) sdk.Coins
 }
+
+type AccountedPoolKeeper interface {
+	GetAccountedBalance(sdk.Context, uint64, string) math.Int
+}
```
amityadav0, line 85: I don't think we are using it, if not let's remove this

amityadav0: got it, so we use to pass values, why not use `k.ExitPoolEst` directly ? we can avoid this extra dep here

amityadav0: `k.ammKeeper.ExitPoolEst`

app/app.go
```diff
@@ -1013,6 +1013,7 @@ func NewElysApp(
 		app.CommitmentKeeper,
 		app.AssetprofileKeeper,
 		app.MasterchefKeeper,
+		app.AccountedPoolKeeper,
```
amityadav0, line 1016:
```suggestion
```

testutil/keeper/leveragelp.go
```diff
@@ -43,6 +43,7 @@ func LeveragelpKeeper(t testing.TB) (*keeper.Keeper, sdk.Context) {
 		nil,
 		nil,
 		nil,
+		nil,
```
amityadav0, line 46:
```suggestion
```

x/leveragelp/keeper/position.go
```diff
@@ -387,27 +388,33 @@ func (k Keeper) GetPositionsForAddress(ctx sdk.Context, positionAddress sdk.Addr
...
-		if err != nil {
-			return sdk.ZeroDec(), err
-		}
-		positionVal = positionVal.Add(sdk.NewDecFromInt(exitCoinsAfterExitFee.AmountOf(depositDenom)))
+		leveragedLpAmount = leveragedLpAmount.Add(commitment.Amount)
+	}
+
+	exitCoinsAfterFee, _, err := k.amm.ExitPoolEst(ctx, position.GetAmmPoolId(), leveragedLpAmount, baseCurrency)
+	if err != nil {
+		return sdk.Dec{}, err
```
amityadav0, line 410: I think this should be `sdk.ZeroDec()` instead of `sdk.Dec{}`

amityadav0 (approved):
Looks good

cosmic-vagabond (approved):

PR #692, opened 2024-08-04
[LeverageLp]: Liquidation algorithm

amityadav0:
## Description

- Add new liquidation algorithm
- Delete old functions
- Delete corrupted data
- Add new algorithm for interest calculation in stablestake module
- Add tests

### What has Changed?

What specific problem were you aiming to address, and how did you successfully resolve it? If tests were not uploaded for this pull request or if coverage decreased, please provide an explanation for the change.

---

### Author Checklist

_All items are required. Please add a note to the item if the item is not applicable and
please add links to any relevant follow up issues._

I have...

- [ ] included the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [ ] added `!` to the type prefix if API or client breaking change
- [ ] targeted the correct branch (see [PR Targeting](https://github.com/elys-network/elys/blob/main/CONTRIBUTING.md#pr-targeting))
- [ ] provided a link to the relevant issue or specification
- [ ] followed the guidelines for [building modules](https://github.com/elys-network/elys/blob/main/docs/docs/building-modules)
- [ ] included the necessary unit and integration [tests](https://github.com/elys-network/elys/blob/main/CONTRIBUTING.md#testing)
- [ ] included comments for [documenting Go code](https://blog.golang.org/godoc)
- [ ] updated the relevant documentation or specification
- [ ] reviewed "Files changed" and left comments if necessary
- [ ] confirmed all CI checks have passed

### Reviewers Checklist

_All items are required. Please add a note if the item is not applicable and please add
your handle next to the items reviewed if you only reviewed selected items._

I have...

- [ ] confirmed the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [ ] confirmed `!` in the type prefix if API or client breaking change
- [ ] confirmed all author checklist items have been addressed
- [ ] reviewed state machine logic
- [ ] reviewed API design and naming
- [ ] reviewed documentation is accurate
- [ ] reviewed tests and test coverage
- [ ] manually tested (if applicable)

### Deployment Notes

Are there any specific considerations to take into account when deploying these changes? This may include new dependencies, scripts that need to be executed, or any aspects that can only be evaluated in a deployed environment.

### Screenshots and Videos

Please provide any relevant before and after screenshots by uploading them here. Additionally, demo videos can be highly beneficial in demonstrating the process.


x/leveragelp/keeper/position.go
```diff
@@ -312,6 +242,8 @@ func (k Keeper) GetPositions(ctx sdk.Context, pagination *query.PageRequest) ([]
 		var position types.Position
 		err := k.cdc.Unmarshal(value, &position)
 		if err == nil {
+			debt := k.stableKeeper.GetDebt(ctx, position.GetPositionAddress())
+			position.Liabilities = k.stableKeeper.GetInterest(ctx, debt.LastInterestCalcBlock, debt.LastInterestCalcTime, debt.Borrowed.ToLegacyDec())
```
avkr003, line 246: Why can't we fetch the liabilities separately wherever its required. this will slow down queries where it's not needed

x/leveragelp/types/codec.go
```diff
@@ -40,6 +41,9 @@ func RegisterInterfaces(registry cdctypes.InterfaceRegistry) {
 	registry.RegisterImplementations((*sdk.Msg)(nil),
 		&MsgAddCollateral{},
 	)
+	registry.RegisterImplementations((*sdk.Msg)(nil),
+		&MsgClosePositions{},
```
avkr003, line 45: Can put this line as well as line 42 below lines 38.

x/leveragelp/types/message_close_positions.go
```diff
@@ -0,0 +1,47 @@
...
+	bz := ModuleCdc.MustMarshalJSON(msg)
+	return sdk.MustSortJSON(bz)
+}
+
+func (msg *MsgClosePositions) ValidateBasic() error {
+	_, err := sdk.AccAddressFromBech32(msg.Creator)
+	if err != nil {
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidAddress, "invalid creator address (%s)", err)
+	}
+	return nil
```
avkr003, line 46: Add validation for checking Liquidate and StopLoss as well. The Address in them is of string type. Need to validate if valid address.

x/leveragelp/types/message_close_positions_test.go
```diff
@@ -0,0 +1,40 @@
...
+			err: sdkerrors.ErrInvalidAddress,
+		}, {
+			name: "valid address",
+			msg: MsgClosePositions{
+				Creator: sample.AccAddress(),
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
```
avkr003, line 31: test cases for invalid address in PositionRequest for Stop loss and Liquidate

x/stablestake/keeper/debt.go
```diff
@@ -33,6 +34,99 @@ func (k Keeper) UpdateInterestStackedByAddress(ctx sdk.Context, addr sdk.AccAddr
...
+		bz = store.Get(currentBlockKey)
+		endInterestBlock := types.InterestBlock{}
+		k.cdc.MustUnmarshal(bz, &endInterestBlock)
+
+		totalInterest := endInterestBlock.InterestRate.Sub(startInterestBlock.InterestRate)
+		numberOfBlocks := ctx.BlockHeight() - int64(startBlock)
+
+		newInterest := borrowed.
+			Mul(totalInterest).
+			Mul(sdk.NewDec(ctx.BlockTime().Unix() - int64(startTime))).
```
avkr003, line 96: Is it possible to get startTime using startBlock as startInterestBlock should have it's block time.

x/stablestake/keeper/debt.go
```diff
@@ -33,6 +34,99 @@ func (k Keeper) UpdateInterestStackedByAddress(ctx sdk.Context, addr sdk.AccAddr
...
+		if firstStoredBlock > startBlock {
+			bz := store.Get(currentBlockKey)
+			endInterestBlock := types.InterestBlock{}
+			k.cdc.MustUnmarshal(bz, &endInterestBlock)
+
+			totalInterest := endInterestBlock.InterestRate
+			numberOfBlocks := ctx.BlockHeight() - int64(startBlock) + 1
+
+			newInterest := borrowed.Mul(totalInterest).
+				Mul(sdk.NewDec(ctx.BlockTime().Unix() - int64(startTime))).
```
avkr003, line 122: same as above

x/stablestake/keeper/debt.go
```diff
@@ -33,6 +34,99 @@ func (k Keeper) UpdateInterestStackedByAddress(ctx sdk.Context, addr sdk.AccAddr
...
+				Quo(sdk.NewDec(numberOfBlocks)).
+				Quo(sdk.NewDec(86400 * 365)).
+				RoundInt()
+			return newInterest
+		}
+	}
+	params := k.GetParams(ctx)
+	newInterest := borrowed.
+		Mul(params.InterestRate).
+		Mul(sdk.NewDec(ctx.BlockTime().Unix() - int64(startTime))).
```
avkr003, line 132: same as above

avkr003 (changes requested):

x/leveragelp/keeper/position.go
amityadav0: it was required from off-chain bot, we may put it separate but then two queries will be needed, This is fine and more accurate

x/leveragelp/types/codec.go
amityadav0: why ?

amityadav0: okay, you mean to remove additional function call

x/stablestake/keeper/debt.go
amityadav0: yes, we can do that. I added time here for information purpose, so we have time in debt and here too. can use this one too, both are same

x/stablestake/keeper/debt.go
amityadav0: no, we can't do that because we don't have startBlock in this case

x/stablestake/keeper/debt.go
amityadav0: to keep it consistent we use time from debt position

x/stablestake/keeper/debt.go
amityadav0: same as above

x/stablestake/keeper/debt.go
avkr003: startTime here is being received as a function parameter, one has to fetch it outside and send it here, and this function again fetches it in `startInterestBlock`, increasing the number of queries.

x/stablestake/keeper/debt.go
avkr003: We can then fetch it outside the if blocks, and then use it the both

x/stablestake/keeper/debt.go
avkr003: Got it.

avkr003 (dismissed):

proto/elys/leveragelp/tx.proto
```diff
@@ -94,33 +98,39 @@ message MsgUpdateStopLoss {
...
+  uint64 id        = 2;
 }
 
-message MsgRemovePoolResponse {}
\ No newline at end of file
+message MsgRemovePoolResponse {}
+
+message MsgClosePositions {
+  string creator   = 1;
+  repeated PositionRequest liquidate = 2;
```
cosmic-vagabond, line 131: @amityadav0 `PositionRequest` is a type defined under `query.proto` can you move it within the `types.proto` file as it will be used for both query and tx messages from now on, thanks

x/leveragelp/keeper/begin_blocker.go
```diff
@@ -18,6 +18,7 @@ func (k Keeper) BeginBlocker(ctx sdk.Context) {
 
 	if epochPosition == 0 { // if epoch has passed
 		pools := k.GetAllPools(ctx)
+		// TODO: Optimise fallback system by dividing traversal in multiple blocks
```
cosmic-vagabond, line 21: @amityadav0 how does the fallback system works I thought we could disable it with a feature flag, by the look of the code it looks like it will check for unhealthy positions across all the positions and liquidate or stop loss close

proto/elys/leveragelp/tx.proto
amityadav0: ok

x/leveragelp/keeper/begin_blocker.go
amityadav0: we are still discussing it

amityadav0: added flag

avkr003 (approved):

amityadav0:
Please don't merge this yet, I have found some issue in queries due to some changes, fixing that

amityadav0:
Fixed now

x/leveragelp/keeper/begin_blocker.go
```diff
@@ -15,47 +16,41 @@ func (k Keeper) BeginBlocker(ctx sdk.Context) {
...
+		pageReq := &query.PageRequest{
+			Limit:      uint64(params.NumberPerBlock),
+			CountTotal: true,
+		}
+		offset, _ := k.GetOffset(ctx)
+		pageReq.Offset = offset
+		positions, _, err := k.GetPositions(ctx, pageReq)
+		if err != nil {
+			ctx.Logger().Error(errors.Wrap(err, fmt.Sprintf("error fetching paginated positions")).Error())
+			return
```
avkr003, line 31: We should have panic here. Any error that shouldn't happen should directly stop the chain (though it might not happen even with panic)

x/leveragelp/keeper/position.go
```diff
@@ -114,60 +67,6 @@ func (k Keeper) DestroyPosition(ctx sdk.Context, positionAddress string, id uint
 	return nil
 }
 
-// Set sorted liquidation
-func (k Keeper) SetSortedLiquidation(ctx sdk.Context, address string, old sdk.Int, new sdk.Int) {
```
avkr003, line 118: If we have removed this, lets remove the Prefix key of it as well.

x/leveragelp/keeper/position.go
```diff
@@ -114,60 +67,6 @@ func (k Keeper) DestroyPosition(ctx sdk.Context, positionAddress string, id uint
...
-
-		// Add position sort keys
-		addrId := types.AddressId{
-			Id:      position.Id,
-			Address: position.Address,
-		}
-		bz := k.cdc.MustMarshal(&addrId)
-		liquidationKey = types.GetLiquidationSortKey(position.AmmPoolId, position.LeveragedLpAmount, new, position.Id)
-		if len(liquidationKey) > 0 {
-			store.Set(liquidationKey, bz)
```
avkr003, line 142: If we have removed this, let's remove it's Prefix key it as well.

x/leveragelp/keeper/position.go
```diff
@@ -114,60 +67,6 @@ func (k Keeper) DestroyPosition(ctx sdk.Context, positionAddress string, id uint
...
-	}
-	bz := k.cdc.MustMarshal(&addrId)
-	debt := k.stableKeeper.UpdateInterestStackedByAddress(ctx, position.GetPositionAddress())
-	liquidationKey := types.GetLiquidationSortKey(position.AmmPoolId, position.LeveragedLpAmount, debt.Borrowed.Sub(debt.InterestPaid).Add(debt.InterestStacked), position.Id)
-	if len(liquidationKey) > 0 {
-		store.Set(liquidationKey, bz)
-	}
-	stopLossKey := types.GetStopLossSortKey(position.AmmPoolId, position.StopLossPrice, position.Id)
-	if len(stopLossKey) > 0 {
-		store.Set(stopLossKey, bz)
```
avkr003, line 167: same as above

x/leveragelp/keeper/begin_blocker.go
avkr003, line 31: We should have panic here. Any error that shouldn't happen should directly stop the chain (though it might not happen even with panic)

avkr003 (changes requested):

x/leveragelp/keeper/begin_blocker.go
amityadav0: why ? I think this is fine, we will get the log, panic will stop the chain I guess

x/leveragelp/keeper/position.go
amityadav0: it is being used in migration, will be removed after upgrade, I have added comments too

x/leveragelp/keeper/position.go
amityadav0: same

x/leveragelp/keeper/begin_blocker.go
avkr003: Tendermint has panic handlers and doesn't allow the chain to stop unless there is consensus failure (if you get panic inside a tx, the chain doesn't stop but has a whole stack trace in the tx log). The logger here won't print the stack trace, the return undermines the severity of the issue and overall it's better to stop the chain and fix it if there is a critical error that might lead to users losing money.

amityadav0: 1. That's not case here, this is not tx, it is `begin_block`, I have attached a ss for your reference, you can try it on your system as well, (put panic in begin blocker)
Screenshot 2024-08-09 at 12 35 50 PM

2. We have the required log in error message, so we will know this is the point of failure

3. About severity of issue, so this is a fallback system and we won't see error while fetching pages but if that occurs, we need to handle the error, that's why this log is added so we know the source of error

4. We will fix it and it won't panic the chain (that means switching all of the chain resulting in more funds loss)

avkr003: I'll trust your judgment on this matter. In the case of certain chains (such as Cosmos or Osmosis), the BeginBlock function does include a call to panic(err) when necessary. This approach is typically used because the error should always be handled appropriately, and if it cannot be returned correctly, the chain is designed to trigger a panic.

Reference: https://github.com/cosmos/cosmos-sdk/blob/v0.47.11/x/distribution/keeper/allocation.go#L25
The AllocateTokens is called in the BeginBlock of the distribution module. It distributes block rewards to validators

amityadav0: for modules such as bank, distribution etc, these are core modules and necessary to panic if a user don't rewards, question is do we want to halt the chain (stop swaps, perps, etc)because of fallback system not getting paginated queries to liquidate positions ?

amityadav0: okay, so I checked osmosis codebase too, here we have
 https://github.com/osmosis-labs/osmosis/blob/977929f2a4a644318e78b51b510f595e581802e8/x/downtime-detector/abci.go#L11
and 
https://github.com/osmosis-labs/osmosis/blob/977929f2a4a644318e78b51b510f595e581802e8/x/superfluid/abci.go#L11 panic here
and at other places we don't panic, looking at out use case I think we should not panic here

avkr003 (approved):

cosmic-vagabond (approved):
LGTM

PR #697, opened 2024-08-06
Avkr003/minor fixes

avkr003:
This adds exit pool specs, removes redundant code from position and correct module level amino codecs

### DIFF
--- a/x/amm/spec/02_mechanism.md
+++ b/x/amm/spec/02_mechanism.md
@@ -41,8 +41,14 @@ Exiting with a swap requires paying both exit and swap fees.
 
 #### Existing Exit Types
 
-- `ExitPool`
+- `ExitPool`: The exit pool can be done in two ways 
+  1. For a non-oracle pool or an oracle pool with an empty exit token denomination, it's not possible to exit entirely in a specific denomination. When a specified amount of shares needs to be exited, the shareRatio is calculated as follows: 
+           ```shareRatio = exitShares / poolTotalShares``` The exit amount for each pool asset is then determined by multiplying this shareRatio by the respective amount of each asset in the pool.
+  2. For an oracle pool with given exit token denom, exit token amount is calculated by calculating value of each LP token in the pool and then multiplied by exit share amount. 
+     A `weightBreakingFee` is also applied which is calculated if distance from desired weight ratio increases after exiting the pool. `weightBreakingFee = WeightBreakingFeeMultiplier x [ ((weightIn x targetWeightOut) / (weightOut x targetWeightIn)) ^  WeightBreakingFeeExponent ]`
+     where `weightIn = 1 - weightOut`, `weightOut = weight of asset in the pool in terms of value`, `targetWeightIn = 1 - targetWeightOut` and `targetWeightOut = wright of asset in the pool in terms of amount`
 
+In both the methods an `exitFee` is applied before finally transferring it to the user. 
 ### Swap
 
 When swapping assets within the pool, the input token is referred to as `tokenIn` and the output token as `tokenOut`. The module uses the following formula to calculate the number of tokens exchanged:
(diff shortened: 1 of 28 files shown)


x/amm/spec/02_mechanism.md
```diff
@@ -41,8 +41,14 @@ Exiting with a swap requires paying both exit and swap fees.
...
 
-- `ExitPool`
-
+- `ExitPool`: The exit pool can be done in two ways 
+  1. For a non-oracle pool or an oracle pool with an empty exit token denomination, it's not possible to exit entirely in a specific denomination. When a specified amount of shares needs to be exited, the shareRatio is calculated as follows: 
+           ```shareRatio = exitShares / poolTotalShares``` The exit amount for each pool asset is then determined by multiplying this shareRatio by the respective amount of each asset in the pool.
+  2. For an oracle pool with given exit token denom, exit token amount is calculated by calculating value of each LP token in the pool and then multiplied by exit share amount. 
+     A `weightBreakingFee` is also applied which is calculated if distance from desired weight ratio increases after exiting the pool. `weightBreakingFee = WeightBreakingFeeMultiplier x [ ((weightIn x targetWeightOut) / (weightOut x targetWeightIn)) ^  WeightBreakingFeeExponent ]`
+     where `weightIn = 1 - weightOut`, `weightOut = weight of asset in the pool in terms of value`, `targetWeightIn = 1 - targetWeightOut` and `targetWeightOut = wright of asset in the pool in terms of amount`
```
amityadav0, line 49:
```suggestion
     where `weightIn = 1 - weightOut`, `weightOut = weight of asset in the pool in terms of value`, `targetWeightIn = 1 - targetWeightOut` and `targetWeightOut = weight of asset in the pool in terms of amount`
```

x/assetprofile/types/codec.go
```diff
@@ -31,18 +31,18 @@ func RegisterInterfaces(registry cdctypes.InterfaceRegistry) {
 }
 
 var (
-	Amino     = codec.NewLegacyAmino()
-	ModuleCdc = codec.NewAminoCodec(Amino)
+	amino     = codec.NewLegacyAmino()
```
amityadav0, line 34: is there a reason we use `animo` and not `Amino` but `ModuleCdc` ?

x/transferhook/types/codec.go
```diff
@@ -17,6 +22,18 @@ func RegisterInterfaces(registry cdctypes.InterfaceRegistry) {
...
+func init() {
+	RegisterCodec(amino)
+	cryptocodec.RegisterCrypto(amino)
+	sdk.RegisterLegacyAminoCodec(amino)
+
+	// Register all Amino interfaces and concrete types on the authz  and gov Amino codec so that this can later be
+	// used to properly serialize MsgGrant, MsgExec and MsgSubmitProposal instances
+	RegisterCodec(authzcodec.Amino)
+	RegisterCodec(govcodec.Amino)
+	RegisterCodec(groupcodec.Amino)
```
amityadav0, line 38: why do we require these 3 in every module ?

amityadav0: only 1 should be enough, I think ?

x/assetprofile/types/codec.go
avkr003: Just following Cosmos-sdk standard practice. Amino variable should be non-exportable and to use it we should use ModuleCdc
Reference: https://github.com/cosmos/cosmos-sdk/blob/v0.47.11/x/feegrant/codec.go

x/transferhook/types/codec.go
avkr003: These three lines register each module's msgs in the amino codec of authz, gov, and group modules. These 3 modules require messages from all the modules in the chain to marshall and unmarshal the messages when received at their end. Authz to grant permissions and execute msgs, gov to add proposals (from v0.47.x) any tx can be executed using gov proposal and group module I am not aware of but pretty some similar reason

cosmic-vagabond (approved):

PR #707, opened 2024-08-09
Avkr003/estaking keys

avkr003:
Optimizes key prefixes for estaking module.

For secp256k1 type addresses, the earlier key was taking 43 bytes for hrp = "elys" as it was converting bech32 address string to a byte array. Now it directly utilises sdk.AccAddress type which is a byte array of length 20 bytes for secp2556k1 addresses. + 1 byte for length

For secp256r1, it will take 32 bytes + 1 byte for length. (the bec32 address string for this type will be higher than 43 characters)


x/estaking/keeper/elys_stake_change.go
```diff
@@ -7,23 +7,43 @@ import (
...
 
 func (k Keeper) SetElysStakeChange(ctx sdk.Context, addr sdk.AccAddress) {
-	store := prefix.NewStore(ctx.KVStore(k.storeKey), types.KeyPrefix(types.ElysStakeChangeKeyPrefix))
-	store.Set([]byte(addr), addr)
+	store := ctx.KVStore(k.storeKey)
+	key := types.GetElysStakeChangeKey(addr)
+	store.Set(key, addr)
 }
 
 func (k Keeper) GetElysStakeChange(ctx sdk.Context, addr sdk.AccAddress) (found bool) {
```
amityadav0, line 15: this function returns if stake change is present and not the key or value. can we change it's name ?

x/estaking/keeper/elys_staked.go
```diff
@@ -49,3 +51,25 @@ func (k Keeper) GetAllElysStaked(ctx sdk.Context) (list []types.ElysStaked) {
...
+	for ; iterator.Valid(); iterator.Next() {
+		var val types.ElysStaked
+		k.cdc.MustUnmarshal(iterator.Value(), &val)
+		list = append(list, val)
+	}
+
+	return
+}
+
+// remove after migration
```
amityadav0, line 71:
```suggestion
// TODO: remove all legacy prefixes and functions after migration
```

x/estaking/keeper/elys_staked.go
```diff
@@ -1,42 +1,44 @@
...
-
-	b := store.Get(types.ElysStakedKey(address))
-	if b == nil {
-		return math.ZeroInt()
+func (k Keeper) GetElysStaked(ctx sdk.Context, address sdk.AccAddress) types.ElysStaked {
+	key := types.GetElysStakedKey(address)
+	store := ctx.KVStore(k.storeKey)
+	bz := store.Get(key)
+	if bz == nil {
+		return types.ElysStaked{
```
amityadav0, line 23: why do we use this data structure ? key is `address` and we add it again in value ?

amityadav0 (commented):
Thanks
1. Could you please add how much some numbers about space saved ?
2. just to confirm, Is It tested on estaking data ? DS is good after migration ?(I am not sure if snapshot has staking data)

x/estaking/keeper/elys_stake_change.go
avkr003: Here the key and value both are the same. So checking key exists does the same. Didn't want to change the function signature where it is used

x/estaking/keeper/elys_staked.go
avkr003: If we don't it won't get exported in the export genesis and no way to know key while importing genesis

avkr003:
Yes migration is working, though I have doubts why `ElysStakeChange` is not in Genesis but in the KV store. I don't have an understanding of it as of now. So I left it as it is as the focus of this PR is optimize KV store

cosmic-vagabond:
Localnet testing using Testnet data:

```
$ elysd q estaking rewards elys1g3qnq7apxv964cqj0hza0pnwsw3q920lcc5lyg
rewards:
- reward:
  - amount: "37"
    denom: ibc/2180E84E20F5679FCC760D8C165B60F42065DEF7F46A72B447CFF1B7DC6C0A65
  - amount: "64606112"
    denom: ueden
  - amount: "538384699"
    denom: uedenb
  validator_address: elysvaloper1g3qnq7apxv964cqj0hza0pnwsw3q920lsclzek
total:
- amount: "37"
  denom: ibc/2180E84E20F5679FCC760D8C165B60F42065DEF7F46A72B447CFF1B7DC6C0A65
- amount: "64606112"
  denom: ueden
- amount: "538384699"
  denom: uedenb
```

cosmic-vagabond (approved):

PR #709, opened 2024-08-12
Fixing set pool function

avkr003:
Currently, there is no way to add new pools to leverage lp module. This update allows to do add or update pools in leverage lp module


amityadav0:
https://github.com/elys-network/elys/blob/main/x/leveragelp/keeper/msg_server_add_pool.go
There is

amityadav0:
This update is not required

PR #711, opened 2024-08-12
Avkr003/commitment keys

avkr003:
Optimizes keys for commitments. 

For secp256k1 type addresses, the earlier key took 43 bytes for hrp = "elys" as it converted the bech32 address string to a byte array. Now it directly utilises sdk.AccAddress type is a byte array of length 20 bytes for secp2556k1 addresses. + 1 byte for length

For secp256r1, it will take 32 bytes + 1 byte for length. (the bec32 address string for this type will be higher than 43 characters)

Tested for migration with chain upgrade.


cosmic-vagabond:
@avkr003 there is a migration error with the commitment module see https://github.com/elys-network/elys/actions/runs/10352124043/job/28657954291?pr=711#step:21:50

avkr003:
I might not be aware of this, v3 to v4 migration have been done? Looks like not because it's trying to find that.

avkr003:
@cosmic-vagabond fixed.

cosmic-vagabond (approved):

PR #727, opened 2024-08-16
fixing claim rewards cli

avkr003:


x/masterchef/client/cli/tx_cmds.go
```diff
@@ -280,7 +280,7 @@ $ %s tx masterchef claim-rewards --from mykey --pool-ids [pool-ids]
 			poolIds := []uint64{}
 
 			poolIdsString, err := cmd.Flags().GetString(FlagPoolIds)
-			if err != nil {
+			if err == nil && poolIdsString != "" {
```
amityadav0, line 283: could you please add tests too ? Looks like this command was never used

avkr003: There is no unit test for the whole package, and adding that will have to deprioritize leveragelp. This is cli cmd only, so will add it later on.

cosmic-vagabond (approved):

x/masterchef/client/cli/tx_cmds.go
amityadav0: okay, please add a task for it

PR #731, opened 2024-08-19
Removes UserRewardInfo with 0 RewardPending and 0 RewardDebt (masterc…

avkr003:
Tested migration on 1,000,000 UserRewardInfo entries.

Removes UserRewardInfo when 0 rewards and 0 debt when the user does claim reward tx.


amityadav0 (approved):
LGTM, Please test it on localnet and verify data with query

cosmic-vagabond (approved):

PR #755, opened 2024-08-27
feat: add dynamic leverage

cryptokage1996:
Enable users to adjust the leverage of an existing position when adding new collateral, allowing for flexibility in leverage management and improving user experience.

### What has Changed?

- previously user have no option to add collateral with different leverage for an open position, now user can specify different leverage value to open a consolidate position
- add collateral flow is removed. if a user want to add more collateral to improve position health, user need to open consolidate position with leverage=1

PR #762, opened 2024-08-31
[MasterChef] Pools rewards - Eden

fenriz07:
Only EDEN rewards can be given to certain pools and not every pool proposed by the community.


x/masterchef/module_simulation.go
```diff
@@ -23,7 +23,11 @@ var (
 )
 
 const (
-// this line is used by starport scaffolding # simapp/module/const
+	opWeightMsgTogglePoolEdenRewards = "op_weight_msg_toggle_pool_eden_rewards"
+	// TODO: Determine the simulation weight value
+	defaultWeightMsgTogglePoolEdenRewards int = 100
```
Wgil, line 28: anyone can help figure this value before merging please?

Wgil (commented):
@fenriz07 are you comfortable with creating unit tests for this? This PR is decreasing the module coverage by around 5% and creating more work later for @avkr003

x/masterchef/keeper/abci.go
```diff
@@ -236,7 +236,9 @@ func (k Keeper) UpdateLPRewards(ctx sdk.Context) error {
 		}
 
 		// Distribute Eden
-		k.UpdateAccPerShare(ctx, pool.PoolId, ptypes.Eden, newEdenAllocatedForPool.TruncateInt())
+		if pool.EnableEdenRewards {
```
amityadav0, line 239: If we are considering eden rewards for a set of pools, then `newEdenAllocatedForPool` this value needs to be updated too.
In this file we are calculating proxyTVL and proxyTVLTotal, to calculate ratio.
We need to consider this while calculating newEdenAllocatedForPool (exclude or include pool)

x/masterchef/keeper/hooks_user_actions.go
```diff
@@ -56,13 +57,17 @@ func (k Keeper) GetRewardDenoms(ctx sdk.Context, poolId uint64) []string {
 	rewardDenoms := make(map[string]bool)
 	rewardDenoms[ptypes.Eden] = true
 	rewardDenoms[k.GetBaseCurrencyDenom(ctx)] = true
-	keys := []string{ptypes.Eden, k.GetBaseCurrencyDenom(ctx)}
```
amityadav0, line 59: why are we removing this ?

x/masterchef/types/codec.go
```diff
@@ -31,6 +32,9 @@ func RegisterInterfaces(registry cdctypes.InterfaceRegistry) {
 		&MsgUpdatePoolMultipliers{},
 		&MsgAddExternalRewardDenom{},
```
amityadav0, line 33:
```suggestion
		&MsgAddExternalRewardDenom{},
		&MsgTogglePoolEdenRewards{},
```

x/masterchef/types/codec.go
```diff
@@ -31,6 +32,9 @@ func RegisterInterfaces(registry cdctypes.InterfaceRegistry) {
...
 	)
+	registry.RegisterImplementations((*sdk.Msg)(nil),
+		&MsgTogglePoolEdenRewards{},
+	)
```
amityadav0, line 37: we can remove this

x/masterchef/migrations/v5_migration.go
```diff
@@ -0,0 +1,14 @@
...
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+func (m Migrator) V5Migration(ctx sdk.Context) error {
+	pools := m.keeper.GetAllPoolInfos(ctx)
+	for _, pool := range pools {
+		pool.EnableEdenRewards = true
+		m.keeper.SetPoolInfo(ctx, pool)
+	}
```
amityadav0, line 12: this migration will fail, we need to have legacy structure when loading and then new structure when setting.
Please update, for reference you can see [this](https://github.com/elys-network/elys/pull/755/files#diff-0aaa985b1adf1dcd60053a3cc5d0ee45c9bd50e685cd44ee869dd4989be492a3)

amityadav0 (commented):
Please add some screenshots after localnet testing

x/masterchef/module_simulation.go
amityadav0: this is fine, it is autogenerated and is same for other modules too

fenriz07: Yes I sent different unit tests to the new gov proposal. However, I believe that the migrations and simulation shouldn't be part of the coverage.

x/masterchef/keeper/abci.go
amityadav0: 1. [Here](https://github.com/elys-network/elys/blob/3a77b329127f72d76f6b6c02fd7efa855ca72768/x/masterchef/keeper/abci.go#L174) we are calculating proxyTVL considering all pools, so we need to update this value to set of pools we are allowing rewards for
2. The reason is [here](https://github.com/elys-network/elys/blob/3a77b329127f72d76f6b6c02fd7efa855ca72768/x/masterchef/keeper/abci.go#L206), we calculate ratio, so if we want to skip some pools, they should not reflect in proxyTVL

fenriz07:
image

**Evidence**:  This account has liquidity in pools 2 and 3, but just pool 2 generates ueden because that pool is enabled for that.

x/masterchef/migrations/v5_migration.go
amityadav0: is this updated ?

fenriz07: Hello @amityadav0  yes that was uptated. 

x/masterchef/migrations/v5_migration.go

amityadav0: Hey @fenriz07, I don't think this migration will pass because loading logic is based on new position but we want to load legacy, have you tested this script ?

cosmic-vagabond:
@fenriz07 this PR has conflicts

amityadav0 (approved):
LGTM

PR #764, opened 2024-09-02
[LeverageLp]: add position usd value

cryptokage1996:
add a new field in query elys.leveragelp.queryPositionsForAddress for the USD amount of the position

https://elys.atlassian.net/browse/DEV-1695


cosmic-vagabond:
@cryptokage1996 can you share some test results in comment here thanks

amityadav0 (approved):

PR #772, opened 2024-09-04
portfolio keys migration and redundant data deletion

avkr003:


x/tier/keeper/begin_blocker.go
```diff
@@ -1,15 +1,14 @@
...
 
 func (k Keeper) BeginBlocker(ctx sdk.Context) {
-	year, month, day := ctx.BlockTime().Date()
-	dateToday := time.Date(year, month, day, 0, 0, 0, 0, ctx.BlockTime().Location())
-	deleteDate := dateToday.AddDate(0, 0, -8)
+	deleteDate := k.GetDateAfterDaysFromContext(ctx, -8)
 	// Remove last 100 values at each block
-	k.RemovePortfolioLast(ctx, deleteDate.String(), 100)
+	k.RemovePortfolioLast(ctx, deleteDate, 100)
+	// migration does not delete all older entries as we don't know whats last date is there, deleting from past 105 days
```
amityadav0, line 11: I think we can put this logic in migration. will that not work ?

avkr003: Older entries might take a lot of time to finish deleting leading to workflow failure. So deleting in abci.

cosmic-vagabond (approved):

PR #790, opened 2024-09-12
[leveragelp]: fix new add collateral flow 

cryptokage1996:


Wgil (changes requested):
We need tests that confirm the use case we're working with. User should be able to add collateral by going through the consolidate flow.

x/leveragelp/types/msgs.go
```diff
@@ -111,7 +111,7 @@ func (msg *MsgOpen) ValidateBasic() error {
 		return errorsmod.Wrapf(sdkerrors.ErrInvalidAddress, "invalid creator address (%s)", err)
 	}
 	// leverage should be greater than 1
```
Wgil, line 113: This comment doesn't make sense after this update

Wgil:
@avkr003 would you suggest any other test to be created that confirms this use case?

cosmic-vagabond:
@cryptokage1996 could you query the newly created position id `267440` and share the result of the query here thanks

PR #796, opened 2024-09-13
New gov proposal to update the perpetual pool was added

fenriz07:
This new gov proposal has the objective of updating a perpetual pool however we're starting with a single field in order to ensure safety. The field is **enabled**


fenriz07:
Before:
image

After to gov proposal:
image

cosmic-vagabond:
@fenriz07 @amityadav0 @avkr003 instead of adding more messages to perpetual can we make so perpetual works by default with all oracle pools, so that we can get rid of those unnecessary pepertual pools settings

PR #807, opened 2024-09-20
feat: add asset price in position query

cryptokage1996:
Expose price for an asset in position query 
https://elys.atlassian.net/browse/DEV-1889


amityadav0 (approved):
Please resolve a conflict

PR #808, opened 2024-09-20
feat: fix division by zero issue

cryptokage1996:
https://elys.atlassian.net/browse/DEV-1734
Fix division by zero issue


x/amm/types/calc_in_amt_given_out.go
```diff
@@ -68,6 +68,9 @@ func (p Pool) CalcInAmtGivenOut(
 	}
 
 	amountInWithoutSlippage := sdk.NewDecFromInt(tokenOut.Amount).Quo(rate)
+	if tokenAmountIn.IsZero(){
```
Wgil, line 71: So if it's zero, does it mean the user is not getting anything back? This seems to be an edge case for sure. We should still be able to fully close the position though.

Wgil: Would the tx be successful? We still need to make sure the position can be closed.

cosmic-vagabond: @cryptokage1996 we should throw an error that says that the user should input a positive token amount in

amityadav0 (approved):

cosmic-vagabond (approved):

PR #818, opened 2024-09-24
feat: add limit order param

cryptokage1996:
Create a limit param for the number of limit orders that can be opened and define a threshold.
- https://elys.atlassian.net/browse/DEV-76


x/perpetual/types/params.go
```diff
@@ -74,6 +74,7 @@ func NewParams() Params {
 		PoolOpenThreshold:                              sdk.OneDec(),
 		SafetyFactor:                                   sdk.MustNewDecFromStr("1.050000000000000000"), // 5%
 		WhitelistingEnabled:                            false,
+		MaxLimitOrder:                                  (int64)(500),
```
avkr003, line 77: please add validate basic and it's test case for the member

avkr003 (changes requested):

PR #829, opened 2024-09-27
Deleting all negative mtp custody

avkr003:


x/perpetual/keeper/mtp.go
```diff
@@ -430,18 +430,18 @@ func (k Keeper) GetAllLegacyMTP(ctx sdk.Context) []types.LegacyMTP {
...
-		k.cdc.MustUnmarshal(bz, &mtp)
-		newKey := types.GetMTPKey(mtp.GetAccountAddress(), mtp.Id)
-		store.Set(newKey, bz)
-		legacyKey := types.GetLegacyMTPKey(mtp.Address, mtp.Id)
-		store.Delete(legacyKey)
+		bytesValue := iterator.Value()
+		k.cdc.MustUnmarshal(bytesValue, &mtp)
+
+		if mtp.Custody.IsNegative() || mtp.Custody.IsZero() {
+			k.DestroyMTP(ctx, mtp.GetAccountAddress(), mtp.Id)
```
amityadav0, line 443: do we need to update pool balances or return left amount before destroying ?

cosmic-vagabond: @amityadav0 good point, we might need to reset the entire perpetual states to make sure there is no leftovers, but I also believe that the ones that end up negative are because they fail to be destroyed because of a wrong condition that was solved in a previous PR, so we could try that out.

cosmic-vagabond (approved):

PR #833, opened 2024-10-01
Dev 1954 user cannot partially close a position

cryptokage1996:


cosmic-vagabond (approved):
tests results shared in comments

PR #844, opened 2024-10-07
[Perpetual]: Remove payments, update in perpetual pool

amityadav0:
## Description

- keep custody within the perpetual wallet and add it to the custody balance of the position so that it helps improve the position health the same way we do the opposite when it comes to the borrow interest payment that reduces it

### What has Changed?

What specific problem were you aiming to address, and how did you successfully resolve it? If tests were not uploaded for this pull request or if coverage decreased, please provide an explanation for the change.

---

### Author Checklist

_All items are required. Please add a note to the item if the item is not applicable and
please add links to any relevant follow up issues._

I have...

- [ ] included the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [ ] added `!` to the type prefix if API or client breaking change
- [ ] targeted the correct branch (see [PR Targeting](https://github.com/elys-network/elys/blob/main/CONTRIBUTING.md#pr-targeting))
- [ ] provided a link to the relevant issue or specification
- [ ] followed the guidelines for [building modules](https://github.com/elys-network/elys/blob/main/docs/docs/building-modules)
- [ ] included the necessary unit and integration [tests](https://github.com/elys-network/elys/blob/main/CONTRIBUTING.md#testing)
- [ ] included comments for [documenting Go code](https://blog.golang.org/godoc)
- [ ] updated the relevant documentation or specification
- [ ] reviewed "Files changed" and left comments if necessary
- [ ] confirmed all CI checks have passed

### Reviewers Checklist

_All items are required. Please add a note if the item is not applicable and please add
your handle next to the items reviewed if you only reviewed selected items._

I have...

- [ ] confirmed the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [ ] confirmed `!` in the type prefix if API or client breaking change
- [ ] confirmed all author checklist items have been addressed
- [ ] reviewed state machine logic
- [ ] reviewed API design and naming
- [ ] reviewed documentation is accurate
- [ ] reviewed tests and test coverage
- [ ] manually tested (if applicable)

### Deployment Notes

Are there any specific considerations to take into account when deploying these changes? This may include new dependencies, scripts that need to be executed, or any aspects that can only be evaluated in a deployed environment.

### Screenshots and Videos

Please provide any relevant before and after screenshots by uploading them here. Additionally, demo videos can be highly beneficial in demonstrating the process.


x/perpetual/keeper/settle_funding_fee_distribution.go
```diff
@@ -48,48 +42,53 @@ func (k Keeper) SettleFundingFeeDistribution(ctx sdk.Context, mtp *types.MTP, po
...
-	} else {
-		// transfer funding fee amount to mtp address
-		if err := k.bankKeeper.SendCoins(ctx, fundingFeeCollectionAddress, mtpAddress, sdk.NewCoins(fundingFeeAmount)); err != nil {
-			return sdk.Coin{}, err
-		}
+	// update mtp custody
+	mtp.Custody = mtp.Custody.Add(fundingFeeAmount.Amount)
+
+	// update pool custody balance
+	err = pool.UpdateCustody(ctx, mtp.CustodyAsset, fundingFeeAmount.Amount, false, mtp.Position)
```
cosmic-vagabond, line 70: @amityadav0 I believe the boolean should be `true` here for incrementing the custody value rather than `false`, correct me if I am wrong, thanks

x/perpetual/keeper/settle_funding_fee_distribution.go
```diff
@@ -48,48 +42,53 @@ func (k Keeper) SettleFundingFeeDistribution(ctx sdk.Context, mtp *types.MTP, po
...
+	if err != nil {
+		return err
+	}
+
+	// update accounted balance for custody side
+	err = pool.UpdateBalance(ctx, mtp.CustodyAsset, fundingFeeAmount.Amount, false, mtp.Position)
```
cosmic-vagabond, line 76: @amityadav0 the funding fee is collected here https://github.com/elys-network/elys/pull/844/files#diff-943de122eee4295bcf39ae75c877dc118541e577b8cd1ecb4b64ee2cfae9d8e6R62 through out a funding address, we need to either get rid of the funding address and keep track of an accounting at the pool level of all the funding fees that were collected that need to be distributed to the opposite trend or if we want to continue using the funding address then we should send the collected funding fees from this address to the perpetual module wallet and update the custody balance accordingly.

cosmic-vagabond (changes requested):

x/perpetual/keeper/settle_funding_fee_distribution.go
amityadav0: we are sending fees from pool to mtp, so it should be reduced, right ? In fee collection we should increase it as we are collecting

x/perpetual/keeper/settle_funding_fee_distribution.go
amityadav0: got it, I will update to later approach

x/perpetual/keeper/settle_funding_fee_distribution.go
cosmic-vagabond: @amityadav0 nope the fees are collected from mtps to the funding address see here https://github.com/elys-network/elys/pull/844/files#diff-943de122eee4295bcf39ae75c877dc118541e577b8cd1ecb4b64ee2cfae9d8e6R62

here we are dealing with the distribution process, so if any given mtp earns with the fees then the fees should be added to its mtp `custody`

amityadav0: ah, got it

x/perpetual/keeper/settle_funding_fee_distribution.go
amityadav0: after thinking more, I think we should keep it as is,
there will be a edge case when collection address won't have enough to pay to perpetual wallet, although we need to update custody and pool which will not be correct as balance won't match

As of now we handled it using toPay mechanism. I think we should close this PR

cosmic-vagabond: @amityadav0 got it although I believe it would be a better UX if the funding fee is added to the custody amount rather than directly distributed to their wallet, for 2 reasons:

1) we already reduce custody for collecting the funding fee from the traders that follows market trend, so it would actually make sense to add to the custody rather than distribute to their wallet, as we are not using their wallet tokens to pay for funding fees

2) earning funding fees would allow traders to keep their position healthier and therefore afloat, that is more true for positions that go against the market trend.

amityadav0: consider an example:

there are 100 positions, we need to collect from 50 and distribute to 50 as well.
and for collection let's say 10 positions have settled (As we do lazy settlement)

Now if 50 distributions happens, there won't be enough to pay from collection address but we will add it to pool right ?

This may result in error later, when closing positions or any other health related calculations.

Do you have any suggestions on this ?

cosmic-vagabond: agreed the first approach allows to avoid having to move liquidity around that way we could optimize the accounting flow

amityadav0: okay, I am adding that

x/perpetual/keeper/settle_funding_fee_distribution.go
amityadav0: updated

x/perpetual/keeper/settle_funding_fee_distribution.go
amityadav0: updated

cosmic-vagabond:
@amityadav0 much better, @amityadav0 can you also remove the funding address from the perpetual params as we have no use of it anymore thanks

amityadav0:
added

x/perpetual/keeper/settle_funding_fee_distribution.go
```diff
@@ -48,48 +31,59 @@ func (k Keeper) SettleFundingFeeDistribution(ctx sdk.Context, mtp *types.MTP, po
...
+	}
+
+	// update pool custody balance
+	err = pool.UpdateCustody(ctx, mtp.CustodyAsset, fundingFeeAmount.Amount, true, mtp.Position)
+	if err != nil {
+		return err
+	}
+
+	// update accounted balance for custody side
+	err = pool.UpdateBalance(ctx, mtp.CustodyAsset, fundingFeeAmount.Amount, true, mtp.Position)
```
cosmic-vagabond, line 71: @amityadav0 the pool balance should not be incremented or decremented upon fee distribution, the distribution is also incremented to the custody amount nothing else

cosmic-vagabond (approved):

PR #859, opened 2024-10-14
DEV-2046/AMM & DEV-2060/ORACLE

fenriz07:
- Oracle: Get asset price endpoint was fixed.
- Oracle: Get all asset price endpoint was fixed.
- Amm: Pools should not be created with more than two assets


x/amm/types/message_create_pool.go
```diff
@@ -45,6 +45,10 @@ func (msg *MsgCreatePool) ValidateBasic() error {
 		return errorsmod.Wrapf(sdkerrors.ErrInvalidAddress, "invalid sender address (%s)", err)
 	}
 
+	if len(msg.PoolAssets) != 2 {
```
avkr003, line 48: Can use `if len(msg.PoolAssts) > 2 || len(msg.PoolAssets) < 2`
Cover both sides.

avkr003 (changes requested):

avkr003 (approved):

cosmic-vagabond (approved):

PR #867, opened 2024-10-17
perpetual module fixes

avkr003:


app/app.go
```diff
@@ -1024,6 +1025,9 @@ func NewElysApp(
 		app.MasterchefKeeper,
 	)
 
+	// Need to do this otherwise app.PerpetualKeeper.LeverageLpKeeper will be nil
+	app.PerpetualKeeper.LeverageLpKeeper = app.LeveragelpKeeper
```
amityadav0, line 1029: okay, so we can't load leveragelpKeeper before perpetualKeeper ?

app/setup_handlers.go
```diff
@@ -30,34 +30,6 @@ func setUpgradeHandler(app *ElysApp) {
...
-
-				// initiate accounted pools
-				pools := app.AmmKeeper.GetAllPool(ctx)
-				for _, pool := range pools {
-					err := app.AccountedPoolKeeper.InitiateAccountedPool(ctx, pool)
-					if err != nil {
-						panic(err)
-					}
-				}
-
```
amityadav0, line 60: This was added recently(not included in upgrade), are we sure we don't need this ?

x/accountedpool/keeper/accounted_pool_initiate.go
```diff
@@ -1,32 +0,0 @@
...
-	}
-
-	for _, asset := range ammPool.PoolAssets {
-		accountedPool.PoolAssets = append(accountedPool.PoolAssets, asset)
-	}
-	// Set accounted pool
-	k.SetAccountedPool(ctx, accountedPool)
-
-	return nil
-}
```
amityadav0, line 32: why are we removing this ?

x/perpetual/keeper/mtp_health.go
```diff
@@ -0,0 +1,53 @@
+package keeper
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	ammtypes "github.com/elys-network/elys/x/amm/types"
+	"github.com/elys-network/elys/x/perpetual/types"
+)
+
+// GetMTPHealth Health = custody / liabilities
+// It's responsibility of outer function to update mtp.BorrowInterestUnpaidLiability using UpdateMTPBorrowInterestUnpaidLiability
```
amityadav0, line 10: so when doing a query, are we making sure latest health is returned ? (considering accrued, funding rate and interest rate)

amityadav0 (commented):
Just wanted to confirm this, are we settling funding rate and interest payment in each case when amount is changed ?

app/setup_handlers.go
avkr003: I am doing it through migrations. No need to InitiateAccountedPool here or in migration, it gets trigger in hooks in enable pool tx of perpetual

x/perpetual/keeper/mtp_health.go
avkr003: Yes. Need to call UpdateMTPBorrowInterestUnpaidLiability, it updates the value (doesn't set in KV store)
Earlier calling this in GetMTPHealth was causing to calculate interest amount multiple times (2-3 times) thus giving incorrect values

x/accountedpool/keeper/accounted_pool_initiate.go
avkr003: I have simplified the structure, all moved to hokks_amm and hooks_perpetual

app/app.go
avkr003: perpetualKeeper needs leveragelpKeeper
leveragelpKeeper needs perpetualKeeper

avkr003:
> Just wanted to confirm this, are we settling funding rate and interest payment in each case when amount is changed ?

@amityadav0 On consolidate, we accumulate the unpaid interest. On close partial or fully, we always settle the unpaid interest fully

app/app.go
amityadav0: ah, got it. why do we leveragelp in perpetual ? is there a way to avoid cyclic dependency ? (would be better I think)

x/perpetual/keeper/mtp_health.go
amityadav0: got it, thanks

app/app.go
avkr003: Sorry, My bad there are 3 levels of cyclic dependencies, masterchefKeeper needs perceptual, perpetual depends on leverage to enable pools, and leverageLp needs masterchef

cosmic-vagabond: we should absolutely avoid that guys this will cause more problems down the road, similarly to AMM and Perpetual, to avoid mutual dependencies between modules we introduce a new module "accounted pool", and relied on Hooks to avoid cyclic dependencies, can we improve the dependency design here as well @avkr003 cc @amityadav0

avkr003: @cosmic-vagabond I agree with you and I also wanted to avoid this, but the constraint came from the fact only those pools that have been enabled for leverage lp can be enabled for perpetual. Due to time constraint I went ahead with this and improve it later.
My second idea is to automatically enable pools for perpetual which gets enabled for leverage lp through hooks.

cosmic-vagabond: @avkr003 agreed all those are valid points and solutions worth trying, can we just make sure to track those future changes as part of a new ticket so we make sure we can plan around that as well thank you

cosmic-vagabond:
@avkr003 some tests are failing

Wgil (approved):

PR #881, opened 2024-10-24
swap fixes

avkr003:

PR #884, opened 2024-10-25
effective leverage

avkr003:


avkr003:
@cosmic-vagabond please merge this only after you have merged the https://github.com/elys-network/elys/pull/881

PR #890, opened 2024-10-28
adding more details for perpetual query

avkr003:


proto/elys/perpetual/query.proto
```diff
@@ -293,6 +293,15 @@ message PoolResponse {
     (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec",
     (gogoproto.nullable) = false
   ];
+  cosmos.base.v1beta1.Coin totalLiabilities = 11 [ (gogoproto.nullable) = false ];
```
cosmic-vagabond, line 296: @avkr003 please use snake case only for proto field names

x/perpetual/keeper/pool_health.go
```diff
@@ -85,3 +88,37 @@ func (k Keeper) CheckMinimumCustodyAmt(ctx sdk.Context, poolId uint64) error {
...
+	totalLiabilities := math.LegacyZeroDec()
+	for _, poolAsset := range pool.PoolAssetsLong {
+		// for long, liabilities will always be in base currency
+		totalLiabilities = totalLiabilities.Add(poolAsset.Liabilities.ToLegacyDec())
+	}
+
+	tradingAsset := ""
+	for _, poolAsset := range pool.PoolAssetsLong {
+		if poolAsset.AssetDenom != baseCurrency {
+			tradingAsset = poolAsset.AssetDenom
```
cosmic-vagabond, line 109: @avkr003 you can break the loop right after assigning the value to `tradingAsset` as there is no need to continue iterating over the loop afterwards

cosmic-vagabond (changes requested):

PR #920, opened 2024-11-08
[Tier]: New query and endpoint to get user data

fenriz07:
## Description

A new endpoint and query were added to get the data user relationship with:

- All users that are providing liquidity in amm pools including the commitment data.
- All users that are currently staking USDC in amm USDC pool including the commitment data.

### Screenshots and Videos

Captura de pantalla_20241108_115127
Captura de pantalla_20241108_115624


x/tier/keeper/query_get_users_pool_data.go
```diff
@@ -0,0 +1,124 @@
...
+	for _, user := range listCommitments {
+
+		u := types.UserData{
+			User:  user.Creator,
+			Pools: []*types.Pool{},
+		}
+
+		for _, commitment := range user.CommittedTokens {
+			if strings.HasPrefix(commitment.Denom, "stablestake/share") {
+				fiatValue := commitment.Amount.ToLegacyDec().Mul(params.RedemptionRate).Mul(tokenPrice)
```
cosmic-vagabond, line 59: @fenriz07 what are the expected denom values for `commitment.Denom`? if it anything other than `USDC` then the tokenPrice should be retrieved using the `commitment.Denom`:

```
	tokenPrice := k.oracleKeeper.GetAssetPriceFromDenom(ctx, commitment.Denom)
```

right?

cosmic-vagabond: all good token price is actually used to covent pool values from usdc to usd

cosmic-vagabond (approved):

PR #958, opened 2024-11-16
fixing tier keeper pointers

avkr003:


app/keepers/keepers.go
```diff
@@ -621,7 +621,7 @@ func NewAppKeeper(
 		app.LeveragelpKeeper,
 		app.StablestakeKeeper,
 	)
-	app.AmmKeeper.SetTierKeeper(&app.TierKeeper)
+	app.AmmKeeper.SetTierKeeper(app.TierKeeper)
```
amityadav0, line 624: If we set whole value here and if there are changes in tier keeper, will it reflect here ?

avkr003: Yes, I made app.TierKeeper pointer as well

PR #1004, opened 2024-11-28
Fix Join amm pool method

99adarsh:
## Description
- The TokenJoined amount was being added to PoolAssets, but the full user amount (MaxAmountIn) was transferred to the pool account.
- Fixed it by transferring only TokenJoined amount(the amount which joins the pool out of MaxTokenIn) to the pool account.


x/amm/keeper/query_join_pool_estimation_test.go
```diff
@@ -32,6 +32,6 @@ func TestJoinPoolEstimation(t *testing.T) {
 		AmountsIn: sdk.NewCoins(sdk.NewInt64Coin("denom1", 100), sdk.NewInt64Coin("denom2", 200)),
 	})
 	require.NoError(t, err)
-	require.Equal(t, sdk.Coins(resp.AmountsIn).String(), "100denom1,200denom2")
+	require.Equal(t, sdk.Coins(resp.AmountsIn).String(), "100denom1,100denom2")
```
amityadav0, line 35: can we please add a comment here, why this value should be 100 ?
for example: Pool ratio is 1:1, so join pool will accept 100 even though 200 is passed in input

amityadav0 (approved):

PR #1036, opened 2024-12-06
rename ibc module name for autocli fixes

99adarsh:
## Description
- AutoCLI subcommand checks subcommands with module name with Use.
- It fails in case of feeibc, interchainaccounts and transfer module as their use statement is (ibc-fee, interchain-accounts and ibc-transfer)

Results:

image

image


avkr003:
Not an important one for the mainnet release. Other repositories have these issues as well. Will fix it later if needed

PR #1080, opened 2024-12-24
chore: use errors.New to replace fmt.Errorf with no parameters

RiceChuan:
use errors.New to replace fmt.Errorf with no parameters


avkr003:
Adding ref:
https://github.com/googleapis/google-cloud-go/issues/9749

avkr003 (approved):

PR #1105, opened 2025-01-13
Adding single perpetual trigger

avkr003:


x/leveragelp/keeper/position_close.go
```diff
@@ -52,9 +52,11 @@ func (k Keeper) ForceCloseLong(ctx sdk.Context, position types.Position, pool ty
 
 	positionOwner := sdk.MustAccAddressFromBech32(position.Address)
 
+	// TODO This means bot failed to close position on time, need to forcefully close the position
```
amityadav0, line 55: so instead of throwing error here, shouldn't we close with remaining assets ?

x/perpetual/keeper/mtp_trigger_test.go
```diff
@@ -0,0 +1,159 @@
...
+	mtp, err := suite.app.PerpetualKeeper.GetMTP(suite.ctx, positionCreator, mtpOpenResponse.Id)
+	suite.Require().NoError(err)
+	pool, _ = suite.app.PerpetualKeeper.GetPool(suite.ctx, mtp.Id)
+	ammPool, _ = suite.app.PerpetualKeeper.GetAmmPool(suite.ctx, mtp.AmmPoolId)
+	return mtp, pool, ammPool, addr[0]
+}
+
+func (suite *PerpetualKeeperTestSuite) TestMTPTriggerChecksAndUpdates() {
+	mtp, pool, ammPool, _ := suite.resetForMTPTriggerChecksAndUpdates()
+	//baseCurrencyEntry, _ := suite.app.AssetprofileKeeper.GetEntry(suite.ctx, ptypes.BaseCurrency)
```
amityadav0, line 54:
```suggestion
```

amityadav0 (approved):

x/leveragelp/keeper/position_close.go
avkr003: Yes. But have to design it more carefully. So whenever there is trigger where we update stop loss price or any other, we do all necessary checks, such as updating/payment of interests, checking if unhealthy and closing if so.

It's related to leverage lp, so didn't do in this PR

PR #1111, opened 2025-01-20
preventing single token exit from amm

avkr003:


x/amm/types/message_exit_pool.go
```diff
@@ -30,6 +31,10 @@ func (msg *MsgExitPool) ValidateBasic() error {
 		}
 	}
 
+	if msg.TokenOutDenom == "" {
```
amityadav0, line 34:
```suggestion
	if msg.TokenOutDenom != "" {
```

avkr003: @amityadav0 We are preventing exiting in single denom, non empty string will allow that

amityadav0: right, doesn't this condition throw error when TokenOutDenom is empty ?
it should be throw error when it is not empty ?
am I missing something ?

amityadav0 (approved):

PR #1122, opened 2025-01-24
fixing exit pool

avkr003:


x/leveragelp/keeper/msg_server_close.go
```diff
@@ -11,34 +14,43 @@ import (
...
 
-	if k.hooks != nil {
-		ammPool, err := k.GetAmmPool(ctx, closedPosition.AmmPoolId)
-		if err != nil {
-			return nil, err
-		}
-		err = k.hooks.AfterLeverageLpPositionClose(ctx, sdk.MustAccAddressFromBech32(msg.Creator), ammPool)
-		if err != nil {
-			return nil, err
-		}
```
amityadav0, line 31: just to confirm we are calling this hook later, right ?

x/leveragelp/keeper/msg_server_close_positions.go
```diff
@@ -26,37 +24,21 @@ func (k msgServer) ClosePositions(goCtx context.Context, msg *types.MsgClosePosi
...
-				return nil, err
-			}
+			liqLog = append(liqLog, position.Id)
+			ctx.Logger().Error(fmt.Sprintf("Unhealthy Position: Address:%s Id:%d cannot be liquidated due to err: %s", position.Address, position.Id, err.Error()))
 		}
 	}
 
 	// Handle stop loss
-	closeLog := []string{}
+	closeLog := []uint64{}
```
amityadav0, line 41: we won't be able to see the reason for not liquidation if we change it to uint64

x/leveragelp/keeper/msg_server_close_positions.go
```diff
@@ -77,27 +59,14 @@ func (k msgServer) ClosePositions(goCtx context.Context, msg *types.MsgClosePosi
...
+			ctx.Logger().Error(fmt.Sprintf("Stop Loss Position: Address:%s Id:%d cannot be liquidated due to err: %s", position.Address, position.Id, err.Error()))
 		}
 	}
 
-	ctx.EventManager().EmitEvent(sdk.NewEvent(types.EventClosePositions,
-		sdk.NewAttribute("liquidations", strings.Join(liqLog, "\n")),
-		sdk.NewAttribute("stop_loss", strings.Join(closeLog, "\n")),
+	ctx.EventManager().EmitEvent(sdk.NewEvent(types.EventFailedClosePositions,
+		sdk.NewAttribute("liquidations", strings.Trim(strings.Replace(fmt.Sprint(liqLog), " ", ",", -1), "[]")),
+		sdk.NewAttribute("stop_loss", strings.Trim(strings.Replace(fmt.Sprint(closeLog), " ", ",", -1), "[]")),
```
amityadav0, line 69: same for liqLog, we will know th id but won't know the reason as it is logged only, not emitted

amityadav0 (approved):

x/leveragelp/keeper/msg_server_close.go
avkr003: Yes, earlier it was missing for begin block closing so I put it in CheckHealthStopLossThenRepayAndClose

x/leveragelp/keeper/msg_server_close_positions.go
avkr003: Logging the reason in even would be bad as it would take lot of space, it will be there in logger though. We will be able to see it, just won't be easy. But what are chances of this happening.

x/leveragelp/keeper/msg_server_close_positions.go
avkr003: same

PR #1123, opened 2025-01-24
check amm pool balance

avkr003:


proto/elys/stablestake/pool.proto
```diff
@@ -0,0 +1,16 @@
+syntax = "proto3";
+package elys.stablestake;
+
+import "gogoproto/gogo.proto";
+import "cosmos_proto/cosmos.proto";
+import "cosmos/base/v1beta1/coin.proto";
+
+option go_package = "github.com/elys-network/elys/x/stablestake/types";
+
+message Pool {
```
amityadav0, line 10: I think we should change this name, borrowedPool or something else, Pool is not good as when we extend this functionality it will be confusing

avkr003: Yeah, we use this structure for generalized lp pool as well, can add extra members for it in the same

amityadav0: generalized lp will have one asset per pool

amityadav0: so we can;t use this, we are going to deposit a single asset for a single pool id, as of now current logic is based on two assets per pool, also liabilities pool and deposit pool are two different things

x/leveragelp/keeper/hooks_amm.go
```diff
@@ -8,27 +8,13 @@ import (
...
+	stablestakePool := k.stableKeeper.GetPool(ctx, ammPool.PoolId)
 
 	for _, asset := range ammPool.PoolAssets {
-		if asset.Token.Denom == depositDenom && price.MulInt(asset.Token.Amount).LT(leverageLpTvl) {
-			return types.ErrInsufficientUsdcAfterOp
+		for _, liabilties := range stablestakePool.TotalLiabilities {
+			if asset.Token.Denom == liabilties.Denom && asset.Token.Amount.LT(liabilties.Amount) {
+				return types.ErrInsufficientUsdcAfterOp
+			}
 		}
```
amityadav0, line 18: This is wrong, I think we need to subtract current liabilities as that is already exited from amm pool and then check

avkr003: Yeah will do that in exit pool PR

amityadav0 (approved):

proto/elys/stablestake/query.proto
```diff
@@ -20,6 +21,16 @@ service Query {
     option (google.api.http).get =
         "/elys-network/elys/stablestake/borrow-ratio";
   }
+
+  rpc Pool(QueryAmmPoolRequest) returns (QueryAmmPoolResponse) {
```
amityadav0, line 25:
```suggestion
  rpc AmmPool(QueryAmmPoolRequest) returns (QueryAmmPoolResponse) {
```
```suggestion
  rpc Pool(QueryAmmPoolRequest) returns (QueryAmmPoolResponse) {
```

PR #1164, opened 2025-02-18
correcting external liquidity for small asset names

avkr003:


Wgil (approved):

x/amm/types/message_feed_multiple_external_liquidity.go
```diff
@@ -23,8 +23,8 @@ func (msg *MsgFeedMultipleExternalLiquidity) ValidateBasic() error {
 
 	for _, liquidity := range msg.Liquidity {
 		for _, depthInfo := range liquidity.AmountDepthInfo {
-			if err = sdk.ValidateDenom(depthInfo.Asset); err != nil {
-				return err
+			if depthInfo.Asset == "" {
+				return errors.New("asset cannot be empty")
```
amityadav0, line 27: this check was added as we want to make sure we get a denom value, 
later on `GetExternalLiquidityRatio` we use this denom to get entry. If it is not in denom terms that could result in non existent value.

amityadav0 (approved):

PR #1175, opened 2025-02-26
DEV-2529 - Updating ojo price-feeder

fenriz07:


amityadav0 (approved):

avkr003:
Changes already in another PR

fenriz07:
@avkr003 we should merge this change asap because if there are new changes that we need to testing on devnet the chain will be panic. This update fixed a panic error on the price-feeder.

PR #1202, opened 2025-03-17
Adarsh/upfront swap

99adarsh:
## Description
- Added new swapExactAmountIn endpoint for skip API which can be called by whitelised addresses


avkr003 (changes requested):

x/amm/migrations/v13_migration.go
```diff
@@ -0,0 +1,14 @@
+package migrations
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+func (m Migrator) V13Migration(ctx sdk.Context) error {
+	legacyParams := m.keeper.GetLegacyParams(ctx)
```
avkr003, line 8: no need for legacy params, Get Params, then if nothing has to be set, then can leave empty

avkr003 (changes requested):

avkr003 (approved):

PR #1230, opened 2025-04-07
removing map from burner and improving DS

avkr003:


x/burner/types/keys.go
```diff
@@ -1,17 +1,43 @@
...
+import sdk "github.com/cosmos/cosmos-sdk/types"
+
 const (
 	// ModuleName defines the module name
 	ModuleName = "burner"
 
 	// StoreKey defines the primary module store key
 	StoreKey = ModuleName
+
+	LegacyHistoryKeyPrefix = "History/value/"
```
amityadav0, line 12: do we need migration for old data ?

amityadav0 (approved):

x/burner/types/keys.go
avkr003: kept it for just in case

PR #1288, opened 2025-05-20
improving enabling pools for leverage lp and perpetuals

avkr003:


proto/elys/perpetual/tx.proto
```diff
@@ -169,6 +169,8 @@ message MsgUpdateEnabledPools {
   // overwritten).
   string authority = 1 [ (cosmos_proto.scalar) = "cosmos.AddressString" ];
   repeated uint64 enabled_pools = 2;
+  repeated uint64 add_pools = 3;
+  repeated uint64 remove_pools = 4;
```
amityadav0, line 173: how about naming enable/disable pools ?

amityadav0: also enabled pools can be removed, so enable_pools, disable_pools will be left

amityadav0 (changes requested):

proto/elys/perpetual/tx.proto
avkr003: Its in tx msg, so we should never remove anything as deserialization of tx will fail.

having enable/disable pools  will still keep the problem of sending all previous pool while adding / removing a single pool

amityadav0: got it, what I meant for renaming is, keep the logic same just rename add -> enable
remove -> disable.
this is fine as well

we are not actually removing the pool, we are enabling or disabling it, am I missing something?

amityadav0 (approved):

proto/elys/perpetual/tx.proto
avkr003: Got the point, I tend to do that because of the naming its called `enabledPools` , if it would have been `enablePools` then could have use that.

99adarsh (approved):

PR #1299, opened 2025-05-26
Add CLOB queries

99adarsh:


proto/elys/clob/query.proto
```diff
@@ -27,16 +28,66 @@ service Query {
...
+      returns (PerpetualOrderResponse) {
+    option (google.api.http).get = "/elys-network/elys/clob/perpetual-order";
+  }
+
+  rpc AllPerpetualOrder(AllPerpetualOrderRequest)
+      returns (AllPerpetualOrderResponse) {
+    option (google.api.http).get = "/elys-network/elys/clob/all-perpetual-order";
+  }
+
+  rpc CurrentTwapPrice(CurrentTwapPriceRequest)
```
avkr003, line 53: Add query for user perpetual order, it will use OrderOwner to fetch order keys and then list of current open orders

avkr003 (approved):

PR #1335, opened 2025-06-16
Add event to track price after swap from non oracle pool

99adarsh:


x/amm/keeper/update_pool_for_swap.go
```diff
@@ -221,6 +221,11 @@ func (k Keeper) UpdatePoolForSwap(
 		)
 	}
 
+	if !pool.PoolParams.UseOracle {
+		tokenInPrice, tokenOutPrice := k.GetEstimatedTokensPriceFromBestPool(ctx, tokenIn.Denom, tokenOut.Denom)
+		types.EmitSwapPriceChangeEvent(ctx, pool.PoolId, tokenIn.Denom, tokenInPrice.String(), tokenOut.Denom, tokenOutPrice.String())
```
avkr003, line 226: the price should be of the pool in which swap is being done. This can report two prices - both different from the price at which swap was done

avkr003 (changes requested):

avkr003 (approved):

PR #1400, opened 2025-07-28
fixing divisible by 0

avkr003:


x/masterchef/keeper/abci.go
```diff
@@ -22,7 +22,7 @@ import (
 func (k Keeper) BeginBlocker(ctx sdk.Context) error {
 	params := k.parameterKeeper.GetParams(ctx)
 	// convert balances in taker address to elys and burn them
-	if ctx.BlockHeight()%int64(params.TakerFeeCollectionInterval) == 0 && params.EnableTakerFeeSwap {
+	if params.EnableTakerFeeSwap && params.TakerFeeCollectionInterval > 0 && ctx.BlockHeight()%int64(params.TakerFeeCollectionInterval) == 0 {
```
amityadav0, line 25: This case is not possible as param setting will be blocked by gov message, we have zero check there

